Issue #1, opened 2024-02-01
Video slider is messed up


hkirat:
Specially in videos with segments. Overlap is happening.

MysteryHawk17:
Hi @hkirat  Please assign this issue to me.

virajp4:
I would like to take up this issue

MysteryHawk17:
@hkirat  ,I have fixed the issue involving the design of the segments in the slider.
You could check it  [https://github.com/code100x/cms/pull/17](here)

PR #15, opened 2024-02-01
fixed video slider issue

rohitbutbliss:
Fixed #1 

By default the line height was set to 0. I have changed the line height to 1 and now text is not overlapping in slider. I have also added word-wrap property to stop text from overflowing when words are too long.

Screenshot 2024-02-01 at 6 21 58 PM

PR #18, opened 2024-02-01
[Bugfix] Overlapping issue in Video Slider

sahilsuman933:
This PR addresses issue #1.

The video slider was overlapping due to the end segment of the slider being placed at the start of the nth second, rather than at the end.

For instance, consider the following segment data:

```
 const segmentsData = [{
    start:0,
    end:3,
    title:"Test - 1"
  },{
    start:4,
    end:6,
    title:"Test - 2"
  },{
    start:7,
    end:10,
    title:"Final Test"
  }
];

```

In the first segment, where the end time is 3 seconds, the current code begins the segment at the start of the 3rd second. Each second contains 1000ms, and this was causing the segments to overlap.

To resolve this issue, I added 1s to the end segment. Consequently, the segment now starts at the end of the nth second, which corrects the problem.

Current Issue which exists in 100xDevs video player:
image

After Implementation:
https://github.com/code100x/cms/assets/34382211/fdec7ea1-be8f-4abe-8f58-471e24b2943a

Note: When providing the segment data, keep in mind that segments shouldn't collide at the end of time, as this will lead to overlap issues. If you want me to add an additional method to ensure this end time does not collide with another segment and is within the video duration, please let me know and I will add those checks.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -70,7 +70,7 @@ export const createSegmentMarkerElements = (
   lastSegment: Segment,
 ) => {
   const segmentEnd = isFinite(segment.end)
-    ? (segment.end / (player.duration() || lastSegment.end)) * 100
+    ? ((segment.end + 1) / (player.duration() || lastSegment.end)) * 100
     : 100;
 
   // Create gap element


sahilsuman933:
Could you please have a look at this PR @hkirat  and share your opinion? Does it satisfy your needs or do you require something else?

hkirat:
Great start can u add some padding around please? Seems a lil constraint

sahilsuman933:
Is this width sufficient, or should I increase it a little more?
image

siinghd:
@sahilsuman933 the problem is not with segments overlapping from my understanding but with the title of the segment.

if the title is too long it doesn't break and go on the next line. This is due to the fixed with of 200px.
Now you can think that changing the width fixes the issue, sort of yes, but now if you last segment title is too long is will overflow the screen.

Try to make the title go on a new line

sahilsuman933:
You are correct, @siinghd, but this PR addresses a different issue. If you watch the attached video, you'll notice that even though I set the end segment to 3 seconds, it appears before the end of the 3 seconds, which shouldn't happen.


https://github.com/code100x/cms/assets/34382211/67e7f88c-8a39-431e-8343-e51c8cbe1770

hkirat:
Closing in favor of https://github.com/code100x/cms/pull/15

Issue #23, opened 2024-02-01
Conflict between `.eslint` and `.prettierrc`


devsargam:
The following two config files contradict each other so
`npm run format:check`
`npm run lint:check`
will not be successful at the same time so, the whole repo needs to follow a single code style and must be linted in a separate pr
 
https://github.com/code100x/cms/blob/825b5ff4ed36745f086bb737af7659bb9b2952a8/.prettierrc#L2
https://github.com/code100x/cms/blob/825b5ff4ed36745f086bb737af7659bb9b2952a8/.eslintrc#L91

devsargam:
I'll work on #19 after your word

devsargam:
@hkirat sorry for the ping but can this get your attention?

Airbornharsh:
@devsargam if resloved then close it

PR #38, opened 2024-02-02
[Bugfix] Workflow Fix

### LINKED ISSUE #19: Lint workflow doesnt seem to be working
https://github.com/code100x/cms/blob/main/.github/workflows/lint.yml isnt triggering on PRs

sahilsuman933:
fixes #19 #23 

This PR addresses three issues:

1. The requirement is to run the workflow on each pull request made to this repository. To achieve this, we need to use the `push` trigger instead of `pull_request`.
2. It resolves several conflicts between eslint and prettier.
a. It changes single quotes to double quotes.
b. The eslint indentation check is removed as prettier and eslint indent do not work together due to differing indentation styles. Reference: https://stackoverflow.com/questions/56337176/prettier-and-eslint-indents-not-working-together
c. Made semicolon necessary.
3. Format the existing codebase

To review whether this workflow is functioning correctly, you can check here: https://github.com/sahilsuman933/cms/actions/runs/7750601270/job/21137093903 and https://github.com/sahilsuman933/cms/pull/2


sahilsuman933:
Please review this PR @hkirat and share me your feedback

siinghd:
> fixes #19 #23
> 
> This PR addresses three issues:
> 
> 1. The requirement is to run the workflow on each pull request made to this repository. To achieve this, we need to use the `push` trigger instead of `pull_request`.


This is not true, push will run the workflow on every push you make. This can lead to unecessary actions time usages because sometimes you don;t need to run it. 

The issue probably is with 
```
pull_request:
    branches:
      - '**'
```

corrected should be 

```
pull_request:
    branches:
      - '*'
```

Also there are a lot of file changes. I think the only thing needed to be changed is prettier config file.

sahilsuman933:
I've reverted the trigger to 'on pull_request'. You can see the change here: https://github.com/sahilsuman933/cms/pull/3

As I noted in my third point, the additional changes are a result of formatting the existing codebase, as it was previously unformatted.

// @hkirat

hkirat:
I see a lot of conflicts, can u get rid of those?

sahilsuman933:
Yes, there are many merge conflicts. I believe it would be more efficient to create a new pull request with changes only in lint.yml, .prettierrc, and .eslintrc. Once that part is merged, I will create another pull request to format the existing code.

Issue #27, opened 2024-02-01
Add "Show Password" Checkbox for /api/auth/signin


sagarb2003:
I would like to  add a "Show Password" checkbox below the password field in the `/api/auth/signin` endpoint for a better user experience.

<h3>Current State</h3>
<img width="958" alt="image" src="https://github.com/code100x/cms/assets/99736036/e01219c9-bee9-4904-acd0-5b875dc93c59">

<h3>Desired Result</h3>
<img width="167" alt="image" src="https://github.com/code100x/cms/assets/99736036/20a452b1-28f5-4097-a274-692dd21b201d">

krishnagptamcs:
naaah , i think it is okay !

siinghd:
@sagarb2003 That's the default signin page of NextAuth. It can be improved with custom page. More info https://next-auth.js.org/configuration/pages.

PR #42, opened 2024-02-02
fix: Fixes #22 and #27 by adding custom signin page and show password with shadcn

### LINKED ISSUE #22: Signin Button looks like link but there should be different background color and text decoration of button have to be removed for better UX
Here is the issue , what it looks like when visiting auth/signin route of 100xdev webpage.
image

shreyas0924:
Fixes #22 and #27 
1. Added custom login page by adding /sigin route and add the same in next-auth pages (auth.ts file)
2. Show and hide password

https://github.com/code100x/cms/assets/72003665/be0eda2f-69bf-41eb-8ae7-e829fee2bed3


src/app/signin/page.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const password = useRef('');
+  const handleSubmit = async (e: React.FormEvent<HTMLButtonElement>) => {
+    e.preventDefault();
+    const res = await signIn('credentials', {
+      email: email.current,
+      password: password.current,
+      redirect: false,
+    });
+
+    if (!res?.error) {
```
siinghd, line 27: I think it would be also better to add an alert or something to notify the user about the error.

siinghd (changes requested):
Add an alert to notify the user about the error

src/app/signin/page.tsx
shreyas0924: I'll do it now

shreyas0924: Added Sonner toast..

image

src/app/signin/page.tsx
```diff
@@ -26,6 +28,13 @@ const Signin = () => {
 
     if (!res?.error) {
       router.push('/');
+    } else {
+      toast('Error Signing in', {
```
siinghd, line 32: Can't this error be more specific? res?.error probably. Might be good to display the returned error insted of static message.

siinghd (changes requested):
Change static message to dynamic

src/app/signin/page.tsx
hkirat: Yeah should address this in a follow up, merging for now as a v1

PR #34, opened 2024-02-02
Done #14(Sticky sidebar) #3(Chapters, this one still need some improvements aka thumbnails), #32 hamburger in appbar

siinghd:
This pr fixes the issues: #32 #3 #14 

Set sidebar to be  sticky at the top on scroll
Added chapters, missing thumbnails (maybe in future)
Moved hamburger to the AppBar instead of body

Preview:

https://github.com/code100x/cms/assets/38940603/e794e2e3-9c84-4b8c-8b23-d721b9cc88d3

Build:

Working on local

PR #46, opened 2024-02-02
Fix urgent #45

### LINKED ISSUE #45: Login broke
https://github.com/code100x/cms/commit/116bd23a88d58a0f33bcf9bd09ec62ce0c4f3791 broke login

email isn't reaching the backend correctly. Have rolled back prod but should be fixed ASAP

siinghd:
Fix #45 

Looking at https://github.com/code100x/cms/blob/main/src/lib/auth.ts, Backend requires `username` field and not `email` as json payload.

This also has #34


shreyas0924:
@siinghd  Sry, I did a silly mistake. Hope ur solution fixes the code

shreyas0924 (changes requested):
Is it possible to add a check that if the user is already authentcated, then redirect to home page.

```
 if(session?.data?.user){
    router.push('/')
   }
```

siinghd:
> Is it possible to add a check that if the user is already authentcated, then redirect to home page.
> 
> ```
>  if(session?.data?.user){
>     router.push('/')
>    }
> ```

Done, moved current signin process to seperate client component. This way the page is rendered on server and can check the session for server redirect instead of client redirect

shreyas0924:
> > Is it possible to add a check that if the user is already authentcated, then redirect to home page.
> > ```
> >  if(session?.data?.user){
> >     router.push('/')
> >    }
> > ```
> 
> Done, moved current signin process to seperate client component. This way the page is rendered on server and can check the session for server redirect instead of client redirect

Good one!!

Neon-20:
Why are we not handling everything inside a middleware.ts file , that would be a cleaner approach. All the protected routing specific things. ref: https://github.com/Neon-20/Authify/blob/ec47c66a90b1115c9d7f209933ef7ae3ebf3f542/middleware.ts (Something like this)

PR #61, opened 2024-02-03
Fix - video player width responsiveness

### LINKED ISSUE #58: Video Player Size does not fill the width when the screen width is reduced.
image

Earlier this wasn't an issue but I think the ideal solution is for it to fill the screen width available.

tanishq-arya:
Fixes #58 
Added full width to the flexbox containing video player and other div(s).


tanishq-arya:
### Before
Screenshot (3)

### After
Screenshot (2)

nimit9:
Fixed this in #64

hkirat:
already fixed

PR #63, opened 2024-02-03
Fix issue 62 - layout shift while using dropdown menu

Lokendra-sinh:
Fixes issue 62 - Radix DropDownMenu Adds Additional Margin/Padding on the body causing layout shift.

[Radix_DropDown_Fixed.webm](https://github.com/code100x/cms/assets/138028282/ac286e7a-d472-4df4-8001-34581e9d879a)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6642 @@
+{
+  "name": "cms",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "cms",
+      "version": "0.1.0",
+      "hasInstallScript": true,
+      "dependencies": {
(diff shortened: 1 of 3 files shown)


hkirat:
Sorry I dont understand the problem that this solves

hkirat:
Closing for noew please re-open if u can elaborate more

Lokendra-sinh:
sure @hkirat 

When the "theme toggler" dropdown menu is not active, you see the scrollbar which is fine

image


But when the dropdown menu is active, the scrollbar disappears and you will observe a slight shift (17px) on the app navbar to the right because the radix component overrides default styles and add additional 17px of margin-right

image

Can you confirm if this is the actual behaviour on your end as well? 

https://github.com/radix-ui/primitives/issues/346#issuecomment-1121440376

Issue #74, opened 2024-02-05
Control Bar Design Improvement


MohitGupta12:
### Proposed Changes
1. Introduce a settings icon to access video quality and playback speed options.
2. Move the progress bar above the control bar for improved visibility.
3. Apply a gradient background to enhance the control bar's aesthetics.

These changes will make a sleeker and more user-friendly control bar.

### Visuals:
**- Before Changes**
<img src="https://github.com/code100x/cms/assets/70692495/2c68fab2-a222-4e53-b563-d830bffb07e8" alt="Before Image" width="600" height="400">

**- After Changes**
<img src="https://github.com/code100x/cms/assets/70692495/682bb81b-6d4a-4708-b76e-24382bdcf6af" alt="Before Image" width="600" height="400">

### Request for Assignment :
Can I go ahead with implementing these changes? If possible, could someone (@hkirat) assign this enhancement to me?

sahilsuman933:
Honestly, it looks pretty good. Feel free to create a PR for it. I'm confident that @hkirat  won't mind enhancing the cleanliness of the video player UI.

sahilsuman933:
Honestly, it looks pretty good. Feel free to create a PR for it. I'm confident that @hkirat won't mind enhancing the cleanliness of the video player UI.

MohitGupta12:
@sahilsuman933 i will start working on it 
can you assign this to me .

sahilsuman933:
I don't have the permission to assign you! @MohitGupta12

MohitGupta12:
@hkirat can you assign this to me

MohitGupta12:
> I don't have the permission to assign you! @MohitGupta12

I will just start working on it

sahilsuman933:
It seems @pankaj-2503 has already started working on it. 

// @MohitGupta12

MohitGupta12:
he just added the color to the progress bar, that's fine I can manage the rest, thanks @pankaj-2503

MohitGupta12:
@sahilsuman933 hey i was researching on this issue, so I found out that we were using video js for our video player and video js has various skins but nothing like youtube so should I came out with 2-3 solutions
 
- Create your control bar and hide videojs's control bar
    this will require to code functionality and style
- Overwrite videojs's control bar with our CSS 
    this would bemainly styles
    
what would be better?

PR #78, opened 2024-02-07
feat: updated the video play background color

pankaj-2503:
image
Added background color on the video play bar for better ui . Hence fixes issue #74

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7091 @@
+{
+  "name": "cms",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "cms",
+      "version": "0.1.0",
+      "hasInstallScript": true,
+      "dependencies": {
+        "@auth/prisma-adapter": "^1.0.6",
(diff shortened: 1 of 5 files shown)


sahilsuman933:
Please remove both of the lock files from the commit.

pankaj-2503:
@sahilsuman933  do you mean package-lock.json file . I thought it would be required so installed is previously.

sahilsuman933:
Yeah. Please remove the `yarn.lock` and `package-lock.json` from the commit. Also, you don't need dotenv package as by default nextjs supports it. So, please remove changes related to that as well.

Docs for reading about environment variables: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables

pankaj-2503:
> Yeah. Please remove the `yarn.lock` and `package-lock.json` from the commit. Also, you don't need dotenv package as by default nextjs supports it. So, please remove changes related to that as well.
> 
> Docs for reading about environment variables: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables
> 
> // @pankaj-2503

I have updated the PR as you said earlier.

PR #79, opened 2024-02-07
bug fix : fixed routing for harnoor and ml course . 

NishantKDash:
The routes for harnoor and ml courses were redirecting to landing page.


hkirat:
We dont have these courses on the platform right now which is why they redirect to https://harkirat.classx.co.in/

Issue #62, opened 2024-02-03
Radix DropDownMenu Adds Additional Margin/Padding on the body causing layout shift


Lokendra-sinh:
## Description: 

Toggling the theme provider component results into additional margin/padding on the document/body causing layout shift.

## Steps to reproduce:

(1) Visit https://app.100xdevs.com/
(2) Toggle the theme 

## Actual Behaviour: shifts the entire app layout resulting into bad user experience

[RadixUi_DropDown_Menu_Issue.webm](https://github.com/code100x/cms/assets/138028282/b92f1241-598a-47be-a105-eed64b22ac11)

Lokendra-sinh:
Looks like when the dropdown menu is active it adds additional margin to the document/body to prevent scrolling.

![image](https://github.com/code100x/cms/assets/138028282/d75fe26a-afb1-40ec-96ab-6a8b9e71114d)


By introducing "modal={false}" in the DropDown root, we can force the component to not apply typical modal styling's and behaviours. This also fixes the issue.

![image](https://github.com/code100x/cms/assets/138028282/24eb3f31-5f68-4910-ac40-626ac54f4697)

Sumit-Kumar-21:
this issue not reflect when i check from my side

PR #83, opened 2024-02-08
Fix: Added modal property to false so that scrolls doesn't occur which was shifting the components when changing theme

pankaj-2503:
Added modal to false in react so that vertical scroll doesn't occur on screen which was shifting the components previously. After doing so it seems everything is as it's position.
Hence fixes issue #62 .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7091 @@
+{
+  "name": "cms",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "cms",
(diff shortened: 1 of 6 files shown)


pankaj-2503:
@sahilsuman933  can you help me why i am getting 1workflow awaiting error  while raising PR.

siinghd:
> @sahilsuman933 can you help me why i am getting 1workflow awaiting error while raising PR.

Kirat need to approve you PR before merging, thats why.

sahilsuman933:
Also, just to mention it is not necessary but a good practice to make a PR title of less than 80 words.

PR #84, opened 2024-02-09
Feat: Changed video playback background color 

pankaj-2503:
image
---
Hence fixes the issue #78


sahilsuman933:
This PR still contains few unnecessary things like `yarn.lock` and dotenv package. Please don't add it on the commit. Instead of using `git add .` use `git add file_name`.

pankaj-2503:
Ok, i understood

PR #88, opened 2024-02-09
Fix deploy error

siinghd:
Issue: https://github.com/vercel/next.js/issues/60301

The error is with the latest version of sharp, i have removed it, it is not needed and should be already present on vercel servers.


siinghd:
@hkirat veere now it should work, tested on https://cms-git-comments-hsingh-siinghd.vercel.app

siinghd:
@hkirat Looks like crashed https://app.100xdevs.com/, any logs on server side?

PR #90, opened 2024-02-10
Added validation for email and password fields

### LINKED ISSUE #89: bug : able to login without credentials (signin field validation missing)
- we can login without entering email and password.

https://github.com/code100x/cms/assets/15179161/b1254cb2-d723-473d-a8e5-e675c660a9ad

- if either email or password is not provided it doesn't present proper error. Faced login issue when the mail and password was auto filled but  sign in request was not sending the username in the request.

https://github.com/code100x/cms/assets/15179161/f6d54fc4-bead-47ef-a6f7-23e8020d1b29

Solution : 
validate if the email or password is not provided.
image
image

TamasRaj:
fixes #89

Added validation for the email and password fields.

https://github.com/code100x/cms/assets/15179161/1d22e3d6-48ae-4f8f-b1d5-cccd15de3bd6

### DIFF
--- a/src/components/Signin.tsx
+++ b/src/components/Signin.tsx
@@ -11,30 +11,54 @@ import React, { useRef, useState } from 'react';
 import { toast } from 'sonner';
 const Signin = () => {
   const [isPasswordVisible, setIsPasswordVisible] = useState(false);
+  const [error, setError] = useState({
+    email: '',
+    password: '',
+  });
 
   function togglePasswordVisibility() {
     setIsPasswordVisible((prevState: any) => !prevState);
   }
+
+  function validateForm() {
+    const state = {
+      email: '',
+      password: '',
+    };
+    if (!email.current) {
+      setError(() => ({ ...state, email: 'Email is required' }));
+      return false;
+    }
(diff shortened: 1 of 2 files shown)


src/components/Signin.tsx
```diff
@@ -11,30 +11,55 @@ import React, { useRef, useState } from 'react';
...
+  const [error, setError] = useState({
+    email: '',
+    password: '',
+  });
 
   function togglePasswordVisibility() {
     setIsPasswordVisible((prevState: any) => !prevState);
   }
+
+  function validateForm() {
```
devsargam, line 23: maybe you could enforce proper email validation with zod or regex. Since the current implementation only checks if the email and password field is empty or not

devsargam (changes requested):

src/components/Signin.tsx
devsargam: Just an advice. I could be wrong

TamasRaj: you can also login using your phone number and password so we should not validate for a valid mail address.

PR #91, opened 2024-02-11
disble download for VideoPlayer2.tsx

iminoaru:
Disabled Right click on video component so user can't directly download the lectures uploaded to prevent piracy.
(There was no issue for this pull request, i felt this needs to be added so i added and tested it)

### DIFF
--- a/src/components/VideoPlayer2.tsx
+++ b/src/components/VideoPlayer2.tsx
@@ -147,6 +147,21 @@ export const VideoPlayer: FunctionComponent<VideoPlayerProps> = ({
 
     document.addEventListener('keydown', handleKeyPress);
 
+    // Disables the right click for video downloading
+    const handleContextMenu = (event) => {
+      event.preventDefault(); // Prevent the default right-click behavior
+    };
+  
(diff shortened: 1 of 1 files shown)


iminoaru:
@hkirat please look into in if it feels necessary.

devsargam:
@iminoaru I don't think this fix is necessary. Since, app.100xdevs.com uses DRM protection to prevent piracy. I could be wrong tho

devsargam:
Can you please upload a demo video or so regarding this problem

iminoaru:
@devsargam the lectures are in mp4 format, so you can download it using browser tools, you can try by right clicking on any lecture uploaded. 
All I did was disable the right click on video component.

devsargam:
@iminoaru oh maybe. But try doing this in the official app.100xdevs.com. Does it still work there?

devsargam:
@iminoaru alright my bad then sorry for bringing this up?

iminoaru:
@devsargam no problem, it's always good to double check stuff.

PR #103, opened 2024-02-21
fix: sidebar scrollable

chetanguptaa:
made sidebar scrollable
to test it out if we make the height really small you can see it work.

Screenshot (54)


hkirat:
Already merged

PR #109, opened 2024-02-23
Video segments UI changes and auto scroll feature

dhruvilmehta:
Added segments to the bottom of the video instead of right side and segments auto scroll on the current segment.
https://github.com/code100x/cms/assets/68022411/187a3a77-2cf3-4fec-a4cd-1bf69a31311e


hkirat:
dont think this UX is needed atm

PR #120, opened 2024-02-26
fix: able to login without credentials (signin field validation missing)

HusainModiwala:
I have added the check for ensuring email and password are non-emty/not-null. And in case they are null (if user didn't provide them), then added the logic to show error message below respected form fields to ask him for the same, and prevent him from moving ahead in such case.

Error message when fields are left blank.
after fix

Error goes away when user starts typing
after fix2


HusainModiwala:
hi @hkirat pls have a look if this solves the issue or something more is needed

PR #119, opened 2024-02-26
FEAT: single session/login

siinghd:
This pr should add support #101 

Key changes need to be added: 

ENV: set  `JWT_SECRET`, correctly as its being used

DB: user model has new field called `token`
`NEXT_PUBLIC_BASE_URL_LOCAL="http://localhost:3000"` -> set this to the correct one

Middleware: check for the token mismatch to redirect the user to signout. 

Currently with nextauth you cannot logout the user on server side so im just redirecting the user to a page and doing singout in the useeffect.

Also middleware runtime is edge so it doesnt  support db querying directly from there  unless you don't use prisma accelerate and prisma edge client.

Missing: admin can logout the user

PR #121, opened 2024-02-26
fix: add missed migration by hshingh

devsargam:
@siinghd missed to commit a migration from prisma

This pr fixes #119


devsargam:
@hkirat could you please check this?
Probably not the best time to ping you

Issue #124, opened 2024-02-27
User token authentication issue (NEXT_PUBLIC_BASE_URL_LOCAL)


RitikBora:
Hey @hkirat ,
I tried setting up the cms project locally. I managed to login using a user and when I try accessing any course I get an  fetch failed error
for the authentication middleware

I suspect its due to NEXT_PUBLIC_BASE_URL_LOCAL being stated as http://127.0.0.1:8000 in .env.example file and I used it in the env file.
This is being called for the backed user check

When i shifted it to http://localhost:3000, the token verification stared to work for me.

 Is this supposed to be http://127.0.0.1:8000 or its a mistake? Please confirm.

Looking forward to your response.

Thanks,
Ritik

devsargam:
@RitikBora HSingh was responsible for the change. It's http://localhost:3000 and not http://localhost:8000 but it depends on  in which port you serve your next application

RitikBora:
Hey @devsargam !
Yeah I understand that. I was just suggesting that we have it as localhost:3000 in the env.example as by default nextjs is served on 3000.

I copied the .env.example file and ran into the error.  Just a suggestion :)

devsargam:
@RitikBora should be fixed in #112

MangaByteCoder:
I'm able to Login with provided userid and password from Readme but its not fetching the test courses.
![Screenshot (3211)](https://github.com/code100x/cms/assets/160803937/e144cb43-b3e8-43af-b458-50956a8107ea)

devsargam:
@RitikBora you can close this. Fixed

PR #122, opened 2024-02-26
Feat: setup docker

devsargam:
This pr setups:
- Docker and docker compose with hot reloading
- Integrates postgres db in the compose
- Fixes the missed migration

PR #125, opened 2024-02-27
Corrected NEXT_PUBLIC_BASE_URL_LOCAL from 8000 to 3000

Maawan:
Issue #124 While setting up the project locally, I realised the url in .env.example is incorrect due to which app is crashing again and again. This is my first contribution hopefully gets merged


devsargam:
Already fixed in #122 btw

PR #129, opened 2024-02-28
fix: Updated comment toast with correct messages

ssk090:
# Pull Request Summary
- Added Material UI Icons which has outlined icons as well as normal icons
- Updated the toast message for upvoting, downvoting and adding comments
- Removed redundant code and optimized the component for better readability and maintainability
- Wrapped the component with React.memo to prevent unnecessary re-renders if the props haven't changed
- Used useCallback to memoize the handleVote function, preventing the creation of a new function on each render, which can improve performance
image


ssk090:
Hey @siinghd 

these are some of the changes on top of your code. LMK if these are good to go

src/components/comment/CommentVoteForm.tsx
```diff
@@ -1,72 +1,75 @@
...
-    },
-  });
-  const handleUpvote = (e: React.FormEvent<HTMLFormElement>) => {
-    e.preventDefault();
-
-    execute({
-      commentId,
-      voteType: VoteType.UPVOTE,
-      currentPath,
+const CommentVoteForm = memo(
```
siinghd, line 13: i think here also memo is not needed as the props does not change based on any state

src/components/comment/CommentVoteForm.tsx
```diff
@@ -1,72 +1,75 @@
 'use client';
+import React, { useCallback, memo } from 'react';
 import { voteHandlerAction } from '@/actions/commentVote';
 import { useAction } from '@/hooks/useAction';
 import { VoteType } from '@prisma/client';
-import { ThumbsDownIcon, ThumbsUpIcon } from 'lucide-react';
+import ThumbUpOffAltOutlinedIcon from '@mui/icons-material/ThumbUpOffAltOutlined';
```
siinghd, line 6: do we need mui icons or its just another library? lucid was fine i think

src/components/comment/CommentVoteForm.tsx
```diff
@@ -1,72 +1,75 @@
...
-  return (
-    <div className="flex gap-2">
-      <form onSubmit={handleUpvote}>
-        <button
-          className="flex items-center gap-1 text-gray-500 dark:text-gray-400"
-          type="submit"
-        >
-          <ThumbsUpIcon className="w-4 h-4" type="submit" />
+    const UpIcon = upVotes !== 0 ? ThumbUpIcon : ThumbUpOffAltOutlinedIcon;
+    const DownIcon = downVotes !== 0 ? ThumbDownIcon : ThumbDownAltOutlinedIcon;
```
siinghd, line 36: I think this is wrong, here upVotes and downvotes are not of the user but a general count. This way you are assuming that the user already voted the filling the icon. It is wrong because if someone upvoted this would result filled for also other users that didnt upvoted

hkirat:
Adding a library to the codebase without consulting with the team is a bad idea usually
We don't want another styling library in the codebase it's an anti pattern to have both tailwind and mui in the same codebase

src/components/comment/CommentVoteForm.tsx
```diff
@@ -1,72 +1,75 @@
...
-          <span>{upVotes}</span>
-        </button>
-      </form>
-      <form onSubmit={handleDownVote}>
-        <button
-          className="flex items-center gap-1 text-gray-500 dark:text-gray-400"
-          type="submit"
-        >
-          <ThumbsDownIcon className="w-4 h-4" />
+    const handleVote = useCallback(
```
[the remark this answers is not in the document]
ssk090: it makes sense not to use the useCallback hook without memo

src/components/comment/CommentVoteForm.tsx
ssk090: yes, you are correct

src/components/comment/CommentVoteForm.tsx
ssk090: Lucid is good. But I did not find any outlined icons so i added mui icons( which is wrong )

src/components/comment/CommentVoteForm.tsx
ssk090: Yes, you are correct. While testing the code I only thought about upvotes/downvotes value as 0 or 1. My bad.

ssk090:
Got a handful of things to learn. Thank you

PR #134, opened 2024-02-29
fix : User token authentication issue (#124) and pressing enter on the signin page should sign the user #123

### LINKED ISSUE #123: Pressing enter on the signin page should sign the user in
After filling login credentials, the submit button has to be separated clicked on or tabbed on to. The form does not submit just by pressing ENTER key after filling in the password

Maawan:
Issue #123 
How I have resolved it ?
Whenever a user is typing password, if the user presses enter then I am calling handleSubmit function which is resolving the issue.
HandleSubmit was expecting a form event e, which is used to prevent the default behaviour of forms. I have  made that parameter as optional because here we are not using <form> so no need for e.preventDefault 

Issue #124 
I have corrected the url of NEXT_PUBLIC_BASE_URL_LOCAL in .env.example from 8000 to 3000


hkirat:
Can u fix the merge conflicts please

Maawan:
@hkirat I have fixed the conflicts

hkirat:
Thankyou! /bounty $10

Issue #139, opened 2024-03-02
Landing Page Navbar (Mobile)


itssishann:
This **issue** is on **mobile Side..**
When **theme is light** and click on **hamburger** button the **join now** button is not in blue color !
image attached 
![Mobile-error](https://github.com/code100x/cms/assets/73055623/8f4c02e2-3dd3-446d-91ed-b7e6d9187ae0)

naman933:
Able to resolve the issue, do assign this issue to me. It is a simple one

PR #141, opened 2024-03-02
issues with light mode fixed

Abishek-Newar:
There was a issue with   colour of join now in light mode so i have fixed it
BEFORE
Screenshot 2024-03-03 010337
AFTER
Screenshot 2024-03-03 010305


Abishek-Newar:
refernce to #139 . it works or mode switch also

Abishek-Newar:
@hkirat please check this code

PR #143, opened 2024-03-03
Update nav-menu.tsx

Aryam2121:
I've changed the text color to black (text-black) to ensure visibility in light mode. #141


hkirat:
Does this work in dark mode? can u share a screenshot of that?

PR #170, opened 2024-03-09
Fix/Feat: comment counters, actions based on role, pinned comment

siinghd:
Fix: 
Fixed problems with comments counter, on delete the counter wasn't descreasing.

Feat: 
Introducing pinned comments, now admin can pin comments(one per content).
Introduced dropdown for actions on single comment.
Now admin can delete,approve chapter comments directly from the dropdown.
This is thanks to do the #154 , as now we have role based access.

Also improved upvoting and downvoting, if user already upvoted and downvoted the icon is filled now

Screenshot 2024-03-09 at 12 05 55 PM


hkirat (changes requested):

siinghd:
@hkirat hi veere, have removed console.log

nimit9:
Hey @siinghd , I think you missed adding the new migration of 'isPinned' column in Comment Table

Issue #172, opened 2024-03-09
CI pipeline should fail on PR if lint fails


hkirat:

Rash-Hit:
@hkirat let me know .. if other changes are required for this issue

PR #173, opened 2024-03-09
fix-ci pipeline should fail on PR if lint fails - #172

Rash-Hit:
Fixes : #172 

Fail the job if `lint:check` exits with a non-zero code


devsargam:
does this work tho?

Rash-Hit:
@hkirat thank you soo much sir ji

PR #191, opened 2024-03-11
updates favicon

nirban07:
the new favicon is same as hkirat.classx.co
hkirat


hkirat:
Sorry am I missing something how does deleting the original favicon add this new one?
Might be missing something obvious here

nirban07:
if you view the binary it shows the updated image. This maybe confusing because you cannot view the binary straight away.
Here are two things that you can check : 
1. click on the commit -> three dots (on the right side ) -> view file -> shows updated favicon
2. click on favicon branch -> src -> app -> favicon -> view raw -> shows updated favicon 
Hope this helps.

hkirat:
Ah yes my bad! Thankyou

PR #196, opened 2024-03-11
Implement 1-Day Caching for getPurchases API Calls

TanmayDhobale:
## Changes Made
- Added an in-memory caching object to store the results of `getPurchases` calls.
- Implemented logic within `getPurchases` to check for valid cached data before proceeding with API calls.
- Cached data is stored with a timestamp to ensure it's only used for up to 24 hours, after which it's refreshed on the next call.
- This caching mechanism is applied conditionally, so it only caches data when the environment does not rely on the local CMS provider for course information.

## Benefits
- Reduced API Calls: Significantly decreases the number of external API requests, lowering our server load and potential costs.
- Improved Performance: Users will experience faster response times when accessing their purchases, enhancing overall user satisfaction.
- Scalability: Reduces the risk of hitting API rate limits as our user base grows.

Considerations- 
The current implementation is an in-memory cache, which means cached data is reset if the server restarts. For long-term scalability, we may consider using a more persistent caching solution like Redis.
-We'll need to monitor the memory usage on our server to ensure the cache does not grow too large.

### DIFF
--- a/src/utiles/appx.ts
+++ b/src/utiles/appx.ts
@@ -1,3 +1,4 @@
+
 import {
   getAllCoursesAndContentHierarchy,
   getAllVideos,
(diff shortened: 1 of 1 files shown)


TanmayDhobale:
@hkirat  ?????

hkirat:
Please use https://github.com/code100x/cms/blob/main/src/db/Cache.ts
generify this to take the `expiry` as an input

TanmayDhobale:
umm okay cool

TanmayDhobale:
@hkirat  can u pls check #199

hkirat:
Closing in favor of https://github.com/code100x/cms/pull/208

PR #199, opened 2024-03-11
Feature: Dynamic Cache Entry Expiry

TanmayDhobale:
This pull request enhances the existing Cache utility by introducing the capability to set dynamic expiry times fvor each cache entry. Previously the cache expiry time was statically set, limiting the utility's flexibility. With this update, users can specify the expiry time in seconds for each set operation, allowing for more nuanced cache control tailored to specific need.

### DIFF
--- a/src/db/Cache.ts
+++ b/src/db/Cache.ts
@@ -13,7 +13,7 @@ export class Cache {
     return Cache.instance;
   }
 
-  set(type: string, args: string[], value: any, expiryInSeconds: number = 3600): void {
+  set(type: string, args: string[], value: any, expiryInSeconds: number = parseInt(process.env.CACHE_EXPIRE_S || '100', 10)): void {
     const currentTime = new Date().getTime();
     const expiry = currentTime + expiryInSeconds * 1000;
     this.inMemoryDb.set(`${type} ${JSON.stringify(args)}`, { value, expiry });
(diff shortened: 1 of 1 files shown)


src/db/Cache.ts
```diff
@@ -1,50 +1,43 @@
...
   }
 
-  set(type: string, args: string[], value: any) {
-    this.inMemoryDb.set(`${type} ${JSON.stringify(args)}`, {
-      value,
-      expiry:
-        new Date().getTime() +
-        parseInt(process.env.CACHE_EXPIRE_S || '100', 10) * 1000,
-    });
+  set(type: string, args: string[], value: any, expiryInSeconds: number = 3600): void {
```
hkirat, line 16: Should defualt to 
```
new Date().getTime() +
        parseInt(process.env.CACHE_EXPIRE_S || '100', 10) * 1000
```

hkirat (changes requested):
Why so much lint diff?
Was the file not linted correctly before or is it bonked right now?

TanmayDhobale:
@hkirat  ummm significant lint difference could indicate a file was not previously linted according to the current project's linting rules, or the linting configuration has  been updated or corrected can u pls check now

TanmayDhobale:
Let's goooooo

PR #200, opened 2024-03-11
Update appx.ts

Aakash231217:
When a request is made, we check if there's a cached version for that email and whether it was stored within the last day. If so, we return the cached result instead of making a new fetch call. If not, you make the fetch call, store the result along with the current timestamp, and then return the new dat


hkirat:
Closing in favor of https://github.com/code100x/cms/pull/208

PR #208, opened 2024-03-11
#195 Used cache in the API calls

### LINKED ISSUE #195: Add big cache to appx api endpoint for purchases
Add a 1 day cache here so dependency on this api reduces - https://github.com/code100x/cms/blob/main/src/utiles/appx.ts#L61

adexh:
Fixes #195
## Changes

- Added use of the Cache class in the purchase API

### DIFF
--- a/src/utiles/appx.ts
+++ b/src/utiles/appx.ts
@@ -14,11 +14,7 @@ const APPX_BASE_API = process.env.APPX_BASE_API;
 const LOCAL_CMS_PROVIDER = process.env.LOCAL_CMS_PROVIDER;
 
 export async function getPurchases(email: string) {
-  const cacheKeyArgs = [email];
-  if (LOCAL_CMS_PROVIDER) {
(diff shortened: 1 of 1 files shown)


src/utiles/appx.ts
```diff
@@ -91,5 +101,6 @@ export async function getPurchases(email: string) {
       }
     }
   }
+  Cache.getInstance().set('courses', [email], responses, 60 * 60 * 24);
```
hkirat, line 104: only this line and the cache get for this is needed
No need for any LOCAL_CMS_PROVIDER logic that can remain uncached

hkirat (changes requested):
just one comment

adexh:
@hkirat Done Sir.
Removed the logic for LOCAL_CMS_PROVIDER and kept its response uncached.

hkirat:
Thankyou! 
/bounty $15

Issue #216, opened 2024-03-12
Inconsitant UI Update


KadlagAkash:
### Description:
Clicking the "Mark as Complete" button does not instantly update the sidebar checkbox. The UI inconsistency persists until a manual reload is performed.

### Steps to Reproduce:

1. Click on the "Mark as Complete" button.
2. Observe that the sidebar checkbox does not update immediately.
3. Perform a manual reload to see the updated status.

### Expected Behavior:
The UI should reflect the updated status of the task in real time when using the "Mark as Complete" button.

### Screenshots
<h3 align='center'>After Clicking on Button</h3>
<img width="1282" alt="image" src="https://github.com/code100x/cms/assets/100425384/824e2404-f6c4-494c-987d-2943bf509fe1">

<h3 align='center'>After Hard Refresh</h3>
<img width="1282" alt="image" src="https://github.com/code100x/cms/assets/100425384/08ac38e0-01d2-4fa7-b699-1f9805edee9b">

devsargam:
Probably can be solved with `useOptimistic` react hook

krishhh16:
I have confidence that I can solve this one. Here's what I learnt from the codebase:-
The inconsistency lies in the face that the sidebar is only rendered once when the whole page loads data from the backend, what's happening right now is that whenever the loadingMarkAs variable(which is responsible for the Mark as Incomplete text to Mark as complete ) is changing, it's not re-rendering the sidebar to also re-render the check icon, meaning, the check mark is probably working independent of the state of the button, and only checks the value when the data is fetched from the database as to whether the course has been marked done or not

what I am going to do is:-
In the parent component of the page that renders both the sidebar and the course video components to that specific page, I will crate a context, which will store both the function and value of that loadingMarkAs variable through useState, and then, make the sidebar re-render the check mark if there is a change in the variable.

I doubt that always fetching to get whether the course is done is not important since it won't create any kind of severe bug. So, having variables take care of this is much better, and only saving the value in the database in case the user arrives back to see which courses have they done.


@hkirat Please let me know if this approach is good and if I can work on this. Or maybe I can create the pull request for you to check if I implemented this correctly?

PR #221, opened 2024-03-14
fix : #216 Mark as complete button UI sync 

Maawan:
Hi @hkirat , to solve this issue I observed that whenever a user is changing content then the sidebar reload data from backend and the checkbox is checked according to completed or not but the problem is only when a user is on the same page and tried to check or click on markascompleted then it could not see the changes in the sidebar and the main content screen accordingly. If sidebar again reloads the data then it will show the correct checked status.

My Approach
I created a atom for this, if any user is changing checkbox then this atom will hold its value and will reflect on the right side and vice versa. As user changes content page, sidebar again gets data from backend and the role of atom is complete.

I am just a beginner, please suggest for the improvement required

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -66,10 +66,8 @@ export function Sidebar({
     const pathArray = findPathToContent(fullCourseContent, contentId);
     if (pathArray) {
       return `/courses/${courseId}/${pathArray.join('/')}`;
-      
     }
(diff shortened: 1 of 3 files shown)


Maawan:
@devsargam I tried to do it that way but didn't got the solution show sidebar's re-render will be triggered. for re-rendering states are the best things so I tried with it.

PR #222, opened 2024-03-14
fix : #216 Mark as complete button UI sync 

Maawan:
Hi @hkirat , to solve this issue I observed that whenever a user is changing content then the sidebar reload data from backend and the checkbox is checked according to completed or not but the problem is only when a user is on the same page and tried to check or click on markascompleted then it could not see the changes in the sidebar and the main content screen accordingly. If sidebar again reloads the data then it will show the correct checked status.

My Approach
I created a atom for this, if any user is changing checkbox then this atom will hold its value and will reflect on the right side and vice versa. As user changes content page, sidebar again gets data from backend and the role of atom is complete.

I am just a beginner, please suggest for the improvement required  

I have attached the video after solving the issue 

https://github.com/code100x/cms/assets/40340661/bcf9c915-ad33-4848-bb55-dbd9991607e1

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -66,10 +66,8 @@ export function Sidebar({
     const pathArray = findPathToContent(fullCourseContent, contentId);
     if (pathArray) {
       return `/courses/${courseId}/${pathArray.join('/')}`;
-      
     }
   };
- 
 
   const renderContent = (contents: any) => {
     return contents.map((content: any) => {
@@ -248,7 +246,6 @@ function Check({ content , pathCheck } : { content: any  , pathCheck : any}) {
   const [currentPath] = useState(usePathname());
   const [markAsComplete , setMarkAsComplete] = useRecoilState(markAsCompleteAtom);
 
-
   return (
     <>
       <input
@@ -257,10 +254,10 @@ function Check({ content , pathCheck } : { content: any  , pathCheck : any}) {
           setCompleted(!completed);
           handleMarkAsCompleted(!completed, content.id);
           setMarkAsComplete({
-            isValid : true,
-            path : currentPath,
-            isCompleted : !completed
-          })
+            isValid: true,
+            path: currentPath,
+            isCompleted: !completed
+          });
           e.stopPropagation();
         }}
         type="checkbox"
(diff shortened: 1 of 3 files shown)


src/components/Sidebar.tsx
```diff
@@ -58,9 +59,19 @@ export function Sidebar({
     const pathArray = findPathToContent(fullCourseContent, contentId);
     if (pathArray) {
       const path = `/courses/${courseId}/${pathArray.join('/')}`;
+      console.log(`Path is this ${path}`);
```
devsargam: could you remove this log

Maawan: In the commit named removed debug statements I have removed these logs

src/store/atoms/markAsComplete.ts
```diff
@@ -0,0 +1,12 @@
+import {atom} from 'recoil';
+
+interface markAsCompleteParams {
+    isValid : boolean
+    path? : string,
+    isCompleted? : boolean
+}
+
+export const markAsCompleteAtom = atom<markAsCompleteParams>({
```
hkirat, line 9: This approach seems slightly ad-hoc-ish
You're using a single atom for all checkboxes
You probably need an atom family here

hkirat (changes requested):

Maawan:
@hkirat So you mean for every module of the course there should be an atom in the atom-family, atom's value should change on triggering checkbox or markasComplete button ?

hkirat:
yes! each atom in the atomfamily should be identified using the courseId

Maawan:
@hkirat Got your point, implementing it

Maawan:
@hkirat This time I changed the approach and tried using Family .
My approach is, I am creating an atomFamily with the atoms of different different courses. 
And every atom will contains an array of objects in this structure 
[{ id : 2,
   parentId : 1,
   isCompleted : true},
{id : 3,
   parentId : 1,
   isCompleted : true}

]
Here id is the content id and parentId is the folder id which is helpfull to display the percentage completed in the FolderView component.

Whenever from Sidebar or ContentRenderer someone hits marksAsComplete or checks the box, firstly I will make the backend request and then update the state for that specific content id

Is this time my approach fine ?

I am stuck in a situation, in atomFamily using atomSelectors , nextJS is allowing me to use server function in client component. I need to getFullCourseContent function to get the course data and convert into this format [{ id : 2,
   parentId : 1,
   isCompleted : true},
{id : 3,
   parentId : 1,
   isCompleted : true}

] and store in state but next is not letting me allow it 

Any suggestions ?

hkirat:
We need the data to be hydrated from the server for SEO

hkirat:
yeah you cant set state on the server
read https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props

PR #231, opened 2024-03-15
feat: setup admin routes to be admin only

devsargam:
Extends #154 

Now only admins are able to view /admin/* routes


devsargam:
image

PR #237, opened 2024-03-18
Feat/bookmarks

nimit9:
### Added Bookmark Feature #4 

[Screencast from 2024-03-18 05-39-59.webm](https://github.com/code100x/cms/assets/37402791/cab97b00-b58d-4d06-9bad-8b9d02f8fc69)

P.S: This assumes that the video will be added to the folder directly and there would be no nested folders.
 
 ### More features to be added

- [ ] Marking bookmarks as favorite
- [ ] Pagination on bookmark page
- [ ] Filtering bookmakrs based on video
- [ ] Some UI changes

@hkirat, if there is a scene of nested folders, I have a solution for that, but that is complicated and requires to make nested db calls to get the parent ids (folders) of the video.  I have somewhat completed code for that as well, so can work on that quickly if needed.
In that case, UI would look something like this:
Screenshot from 2024-03-17 01-22-36


hkirat:
So, I dont think we need this
We do need a way for people to mark videos as favourite/bookmark them
But bookmarking a timestamp seems over engineered/dont think anyone would use it
Closing for now, but feel free to re-open

nimit9:
Okay, then I'll do the same thing with videos instead of the timestamp. 
please assign the issue to me

Issue #234, opened 2024-03-16
Migrating to Turborepo (incrementally)


aryanprince:
# Summary

To migrate the current repo to a monorepo setup using Turborepo. This will allow us to have multiple different applications (like marketing, Discord Bot, so on) in the same repo, while having shared common packages between each other.

# Pros

1. Improved DX for repetitive tasks such as building, linting, and formatting.
2. Will allow us to keep the main 100x dashboard, (future) Discord Bot, marketing sites, and other apps in the future in one place. Ensuring all apps can share packages across each other (such as various configs and Tailwind for example).
3. Saving both time and money in CI - by leveraging Turborepo's Remote Cache. 
4. Improved local dev experience since build caches can be reused by all contributors.
5. ...and not to mention, every other benefit of having a monorepo fundamentally such as allowing shared packages to be updated locally and seeing live changes on the apps importing shared packages.

# Cons

1. Deployments will need to be updated to new directory inside the monorepo (updating deployment directory to `apps/web` for example)
2. Will need to update the Docker Compose script to ensure it works with the new Turborepo setup.

# The Plan

Make a very minimal migration to Turborepo. By this, I mean we use a 2 step approach to the migration in order to keep minimal diffs and manageable PRs/reviews.

- [ ] Step 1. Move the main Nextjs application as is to a folder like `apps/web` first. This is done to produce as little diffs in the PR as possible, ensuring the first PR is very focused and can be reviewed better.
- [ ] Step 2: Later, refactor common packages like ESLint, TS config, Tailwind, and so on to their respective packages in subsequent PRs later on.

# My Experience

I've incrementally migrated from a single Next.js app to a monorepo setup with Turborepo with 3 different Next.js apps [here](https://github.com/aryanprince/campus-hub).

aryanprince:
@hkirat If this is something on the roadmap, can I be assigned to this as I’ve already initiated work on it? Thanks.

hkirat:
this is great

aryanprince:
## Progess

For now, I have managed to successfully migrate the entire codebase to Turborepo. Everything works well in my fork for the time being ([link here](https://github.com/aryanprince/cms)). However, there are several blockers that need to be addressed before I submit my PR.

## Blockers

There are currently 2 blocking issues/PRs: 

1. There are a large number of conflicting ESLint and Prettier config rules which I believe is being fixed in #211. 
2. There are missing migration files from the last DB schema change (part of #170). #237 seems to include the required missing migration files.

[...]

PR #243, opened 2024-03-18
Feature : Refine Linting and Formatting Configuration

TanmayDhobale:
Updated ESLint and Prettier configs for better code consistency. Includes refined rules and pre-commit hook adjustments to automatically fix format and lint issues, making code contributions smoother.

## Changes

- Updated `.eslintrc` with additional rules for TypeScript, improving code quality and consistency.
- Modified the `.prettierrc` file to refine formatting rules, ensuring that automatic code formatting aligns with our project's standards.
- Enhanced the pre-commit hooks in `.husky/pre-commit` to run both formatting and linting fixes automatically, reducing the likelihood of committing code that doesn't meet our guidelines.

## Impact

- Contributors will experience fewer discrepancies when pushing code, as the pre-commit hooks now more thoroughly ensure code adheres to our standards.
- These updates aim to streamline the development workflow by reducing manual linting and formatting interventions required from contributors.
- By enforcing a consistent coding style, we improve the readability and maintainability of our codebase.

## Testing

- Introduced intentional linting/formatting violations to verify that the pre-commit hooks automatically correct them or block the commit when auto-fixes are not possible.
- Confirmed that the updated configurations do not conflict with our current codebase by successfully committing changes after the updates.

Please review the changes and provide any feedback you might have!  fixed #201

PR #211, opened 2024-03-12
Feature: Integrate Code Quality Tools for Enhanced Consistency

TanmayDhobale:
This pull request introduces ESLint, Prettier, and Husky into our development workflow. The integration of these tools serves to enforce a consistent coding standard across the project, ensuring that all contributions adhere to our defined styles and practices. 

Key Changes:
- ESLint: Setup to analyze the codebase for adherence to our coding standards, helping identify potential errors and code smells.
- Prettier: Configured to automatically format code according to our specified styles, ensuring consistency in the appearance and structure of our code.
- Husky: Employed to trigger ESLint and Prettier checks before committing changes, ensuring that only code that meets our quality standards is added to the repository.

These tools collectively contribute to a more maintainable and error-free codebase, streamlining the development process and enhancing collaboration efficiency.

PR #244, opened 2024-03-18
fix: Fixed ESLint styling rules overriding Prettier styling rules

aryanprince:
# Summary

This fixes the minor issue where ESLint's styling rules were overwriting the Prettier styling rules. Videos below showcase an example.

This PR is part of the Turborepo monorepo migration mentioned in #234.

EDIT: Comment below in this PR ([here](https://github.com/code100x/cms/pull/244#issuecomment-2006243552)) showcases an example of the conflicting styling rules caused before this PR.

# Changes

- [x] Removed 2 of ESLint's conflicting styling rules

# Showcase

## Before Changes

👇 Shows how the `lint:fix` script overrides changes done by `format:fix`

https://github.com/code100x/cms/assets/45622345/71cfdeb2-8e62-450a-8d5a-da72d12576d4

## After Changes

👇 ESLint (`lint:fix`) no longer conflicts with Prettier's (`format:fix`) styling changes now

https://github.com/code100x/cms/assets/45622345/ea8b2486-8556-4804-baf4-5acadb1b93f6

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -68,9 +68,7 @@ rules:
   # stylistic errors
   no-spaced-func: 2
   semi-spacing: 2
-  quotes: [2, 'single']
   key-spacing: [2, { beforeColon: false, afterColon: true }]
-  indent: [2, 2]
   no-lonely-if: 2
   no-floating-decimal: 2
   brace-style: [2, 1tbs, { allowSingleLine: true }]


TanmayDhobale:
@aryanprince how u are u solving this issue just removing ?
```
  quotes: [2, 'single']
  indent: [2, 2
```

aryanprince:
@TanmayDhobale Good question, it was very annoying during dev cus of our Husky pre-commit hooks - and being very annoying to debug as well.

Video below shows an example of one of the conflicts caused by ESLint rule: `indent: [2, 2]`. You can see how running `format:fix` formats all files according to Prettier's conventions and our Prettier config file (`.prettierrc`). But running `lint:fix` then overwrites styling rules according to these 2 ESLint rules I just removed in this PR. 

Adding to this, every time someone makes a PR/commit, [our Husky pre-commit rules](https://github.com/code100x/cms/blob/b7d1ca4fab02becfc6efb7abf4f53f44379a5096/.husky/pre-commit#L4) will first apply Prettier formatting using `format:fix`, and the next step is to lint the whole codebase using `lint:fix`. But then again, on every commit the ESLint styling rules will override any of Prettier's styling rules since both of these scripts run on every commit.

https://github.com/code100x/cms/assets/45622345/4ffad8bf-bff7-4f85-8657-102e14c08cce

Issue is having ESLint do _any_ styling for us since Prettier already does a great job at that. It would be best to remove all of the ESLint styling rules we have to prevent any future conflicts. But the 2 rules I removed here were the only ones causing conflicts in this codebase currently.

TanmayDhobale:
> @TanmayDhobale Good question, it was very annoying during dev cus of our Husky pre-commit hooks - and being very annoying to debug as well.
> 
> Video below shows an example of one of the conflicts caused by ESLint rule: `indent: [2, 2]`. You can see how running `format:fix` formats all files according to Prettier's conventions and our Prettier config file (`.prettierrc`). But running `lint:fix` then overwrites styling rules according to these 2 ESLint rules I just removed in this PR.
> 
> Adding to this, every time someone makes a PR/commit, [our Husky pre-commit rules](https://github.com/code100x/cms/blob/b7d1ca4fab02becfc6efb7abf4f53f44379a5096/.husky/pre-commit#L4) will first apply Prettier formatting using `format:fix`, and the next step is to lint the whole codebase using `lint:fix`. But then again, on every commit the ESLint styling rules will override any of Prettier's styling rules since both of these scripts run on every commit.
> 
>  Showcasing.ESLint.and.Prettier.Conflict.mp4 
> Issue is having ESLint do _any_ styling for us since Prettier already does a great job at that. It would be best to remove all of the ESLint styling rules we have to prevent any future conflicts. But the 2 rules I removed here were the only ones causing conflicts in this codebase currently.

yea bcs i try many times but without changing files code lint problem is still there. then i have to change the file code automatically u can see here ive try to solve this issue in #211 but then u solved , great

devsargam:
@aryanprince @TanmayDhobale do u know about extending eslint with prettier. It might be somewhat better than what we are doing currently

 ref = https://github.com/prettier/eslint-config-prettier

aryanprince:
@devsargam I 100% agree it's a better solution and it's what I gravitated towards at first.

But I instead found out that I could solve the conflicts just by disabling the 2 ESLint styling rules for now, instead of introducing a new dependency (like `eslint-prettier-config`) for it.

@TanmayDhobale Feel free to include `eslint-prettier-config` in your next PR. I would recommend you still open a PR (like the one you closed) as you were planning to add more new features, perhaps after this PR gets merged to ensure minimal conflicts.

TanmayDhobale:
> @devsargam I 100% agree it's a better solution and it's what I gravitated towards at first. That's what I use in most of my personal projects. 
> 
> But I instead found out that I could solve the conflicts just by disabling the 2 ESLint styling rules for now, instead of introducing a new dependency (like `eslint-prettier-config`) for it.
> 
> @TanmayDhobale Feel free to include `eslint-prettier-config` in your next PR. I would still recommend you still open a PR (like #243 which you closed) as you were planning to add more new features, perhaps after this PR gets merged to ensure minimal conflicts.

Not sure let's see

aryanprince:
> But one thing, after removing those 2 options now eslint won't provide rules for indentation in this project.

I agree. But we are better off with ESLint **not** providing rules for indentation, since Prettier will auto format the files anyways - especially with the Husky pre-commit hooks in place.

> This issue solves a big problem but accidentally starts another small one

Could you provide some examples where this could start other problems? Currently, all of our other ESLint styling rules (in `.eslintrc`) don't seem to have any overlap with our Prettier styling rules (in `.prettierrc`) so we should be good.

devsargam:
@aryanprince I don't remember much brother. But I did see some weird behaviour when it was a case for nested ternaries
Example
```ts
true ?
  true ?
     "yes" : "no"
  : "something"
```

TanmayDhobale:
after reviewing our current setup and the discussions around ESLint and Prettier conflicts, I suggest integrating eslint-config-prettier as a more sustainable solution. This approach will systematically disable all ESLint rules that could conflict with Prettier, ensuring a smoother workflow without needing to manually disable specific rules. It not only simplifies our setup but also enhances maintainability and minimizes potential issues down the road. let me know pls @devsargam @aryanprince

devsargam:
@TanmayDhobale sure go on. For now @aryanprince solution seems fine as it's blocking the turoborepo migration

aryanprince:
@devsargam I get you, can totally see this happening. Best case scenario would be to update our ESLint config to use `eslint-prettier-config`, and then run the linter across the entire project to fix any issues. But this will cause merge conflicts with every other open PR.

devsargam:
yeah, linting is always tedious and should happen when the pr count is almost 0 else cause a lot of merge conflicts. Upto then you guys keep comitting with --no-verify flag which wont trigger husky

TanmayDhobale:
Agreed, let’s prioritize minimizing merge conflicts by holding off on sweeping lint changes until our PR count is low. Using --no-verify temporarily seems like the best path forward. We’ll circle back to eslint-config-prettier integration later.

TanmayDhobale:
@devsargam yoo whtats ur discord ? or maybe check ur twitter dm once

PR #246, opened 2024-03-19
Fixes #238 Comment type labels

### LINKED ISSUE #238: Type of  Comment DropDown shows incorrect name on change.
In the comment section the user have option to select and change the type of comment.
image
On change to Intro Comments the drop down value is marked with the code rather than description shown below.
image

This should not happen and On change the value should remain as that of the description shown in the drop down initially.

shreyas0924:
Screenshot 2024-03-19 113404
Screenshot 2024-03-19 113353


hkirat:
always give more context of what the change is in the PR description
closing for now

PR #264, opened 2024-03-23
fix: cources from db

devsargam:
@hkirat check this

### DIFF
--- a/src/utiles/appx.ts
+++ b/src/utiles/appx.ts
@@ -7,6 +7,7 @@ import { authOptions } from '@/lib/auth';
 import { Course } from '@/store/atoms';
 import { getServerSession } from 'next-auth';
 import { Cache } from '@/db/Cache';
+import prisma from '@/db';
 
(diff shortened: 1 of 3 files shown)


devsargam:
Bhaiya email is the thing we are using as phone number

devsargam:
@hkirat

src/utiles/appx.ts
```diff
@@ -63,6 +64,23 @@ export async function getPurchases(email: string) {
     return courses;
   }
 
+  // Check if the user exists in the db
+  const coursesFromDb = prisma.course.findMany({
+    where: {
+      purchasedBy: {
+        some: {
+          user: {
+            email,
```
devsargam, line 73: bhaiya it's an array so need to do some there

devsargam: cant do just with user: {email}

devsargam: @hkirat

PR #265, opened 2024-03-23
fix: changed pre-commit and eslitnrc

ofcljaved:
# PR FOR LINTING ERROR

- I just do yarn install it did changes in yarn.lock file but others changes are WORKING
You can check
AND yess The previoud .eslintrc is very big but doesn't configures with prettier so I change it with the one on  daily-code and remove prettier from it

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,122 +1,15 @@
-parser: "@typescript-eslint/parser"
-
-env:
-  node: true
-  es6: true
-  jest: true
-  mocha: true
-  browser: true
-
-plugins:
-  - "@typescript-eslint"
-
-extends:
-  - eslint:recommended
-  - plugin:@typescript-eslint/eslint-recommended
-  - plugin:@typescript-eslint/recommended
-
-parserOptions:
-  ecmaVersion: 2018
-
-rules:
-  # Ignore Rules
-  strict: 0
-  no-underscore-dangle: 0
-  no-mixed-requires: 0
-  no-process-exit: 0
-  no-warning-comments: 0
-  no-use-before-define: 0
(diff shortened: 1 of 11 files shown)


ofcljaved:
@hkirat

ofcljaved:
I know I have made changes to other file to  but if i won't it will not let me commit the code so
Now it won't fix in the pre-commit it'll just check for it and if it found any error or warning it'll exit with code
and then You have to run 
`npm run lint:fix` 
and 
`npm run format:fix`
Then You'll able to commit if it passes

devsargam:
@ofcljaved you can actually ignore the pre-commit hook with
```
git commit -m 'changes' --no-verify
```

ofcljaved:
@devsargam Oh I didn't knew that, Thanks a bunch And also the issue is with `git add . ` and also we want linting too So I just make sure to check for linting in pre-commit and exit based on the exit code of lint.
If the pre-commit fails  then user has to run two scripts lint:fix and format:fix and then commit the changes this way we don't have to use `git add .` in the pre-commit

devsargam:
hi @ofcljaved IMO running `npm run lint:check` and `npm run format:check` would serve better in our scenario. What do you think?

ofcljaved:
> hi @ofcljaved IMO running `npm run lint:check` and `npm run format:check` would serve better in our scenario. What do you think?

In the pre-commit I'm doing link and format check Which shows warning to the user if anything defies the eslint and prettier
Then user has to fix those issue either manually or  by running `npm run lint:fix && npm run format:fix`
The goal is to let user commit the code only when he pass the lint and format check

Issue #268, opened 2024-03-24
Bug - Course Complete Percentage Disappeared - Because of Latest Change 


Rash-Hit:
See  - 
![Screenshot 2024-03-27 132402](https://github.com/code100x/cms/assets/73944177/8606a03e-3cd9-4146-a110-ea420a6ae3b8)

And i think the reason , is because we are not calculating stats again while doing this .. we just fetching and setting in the cache.
![image](https://github.com/code100x/cms/assets/73944177/9c480cd2-b73a-4992-9251-a09d09526bb2)

not able to produce this issue on local because of cms local provider check , but i might be wrong ,

PR #269, opened 2024-03-24
-fix-percentage-issue

Rash-Hit:
See  - 
image

And i think the reason , is because we are not calculating stats again while doing this .. we just fetching and setting in the cache.
image

not able to produce this issue on local because of cms local provider check , but i might be wrong ,

@hkirat  , i might be wrong .. but please check from your side one 

fixes: #268

### DIFF
--- a/src/utiles/appx.ts
+++ b/src/utiles/appx.ts
@@ -75,11 +75,46 @@ export async function getPurchases(email: string) {
         },
       },
     },
+    select: {
+      id: true,
+      title: true,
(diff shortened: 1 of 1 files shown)


Rash-Hit:
@hkirat  check this issue , no percentage completed in the 0-100

hkirat:
yeah we'll be getting rid of caches soon
I'm assuming this resolves after the cache clears (1 day)
Closing for the same reason

PR #275, opened 2024-03-25
fix: the edge case problem

### LINKED ISSUE #272: bug(edge case): Unable to see some courses
### User courses are stored in db so if anyone already in db purchases a new course db is not updated.

ref:https://github.com/code100x/cms/blob/c5c720cd71adff7ada1e2bb6f89f43f38667208d/src/utiles/appx.ts#L68-L87

context:
image

@hkirat could u add this mail to db: `rahul.mymail1@gmail.com`

devsargam:
@hkirat make sure to test the code since I couldn't access appx api
Fixes #272

image

### DIFF
--- a/src/actions/refresh-db/index.ts
+++ b/src/actions/refresh-db/index.ts
@@ -1,0 +1,66 @@
+'use server';
+import db from '@/db';
+import { Cache } from '@/db/Cache';
+import { getAllCourses } from '@/db/course';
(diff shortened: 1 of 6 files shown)


hkirat:
Yeah great hack for now. Was thinking along the same lines. Merging for now but will test tom and push
/bounty $25

Issue #267, opened 2024-03-24
Video Refreshes  when we close or open Chapters and  Playback rate also goes back to 1X


vijayindalkar:
Steps to reproduces 

earlier this doesn't used to happen 

![image](https://github.com/code100x/cms/assets/61607592/32bce5dc-2514-4da9-a766-7ec2b0233050)
 
![image](https://github.com/code100x/cms/assets/61607592/4479129e-b986-4656-9a0a-e25771acb36b)

PR #276, opened 2024-03-26
Update VideoPlayer.tsx

Aryam2121:
The handleChapterChange function is added to handle the chapter change event and refresh the video player accordingly. The initializePlayer function sets the initial playback rate to 1X and initializes the video player with the specified attributes. The useEffect hook is used to trigger the initialization of the video player when the mpdUrl or subtitles prop changes. problem solve of #267


Aryam2121:
Sir check this also >

PR #280, opened 2024-03-27
Refresh DB section update

punyakrit:
It would show only if no course is visible if courses are visible it wont show in main page
Screenshot 2024-03-27 at 11 59 26 AM


punyakrit:
But still why show a unwanted pop up of text in main course page only show if course list are not fetched from db

punyakrit:
lets see if it is a valid contribution @hkirat

hkirat:
@devsargam is right
It is for busting the cache

PR #289, opened 2024-03-29
Feat : Introducing Breadcrumbs for Enhanced Course Navigation and better UX.

### LINKED ISSUE #282: Feature : Implemented dynamic breadcrumbs for course navigation and better UX
### **Problem Statement**  : 
Right now, it's tough for users to know where they are in a course because there's no clear indication. This especially becomes a problem after they've been clicking around in the sidebar.

### **Solution** : 
We can have something called Breadcrumbs , which will show up above the course content. These breadcrumbs will change as users move around, showing exactly where they are in the course. Making it easy to see which part of the course you're in and navigate between sections by clicking, even if the sidebar is closed, for a better user experience.

### Approach I used :
I have used shadcn for breadcrumb component and to generate the breadcrumb based on where the user is , I have used recursive function which checks matched content from full course content and then it pushes to some array and returns it and finally we can see our generated breadcrumbs .

**Reason :**  I am using recursive approach because as of now cms course structure is almost like : course(ex : 0-100) >folders (Week 1 , Week 2 etc.) > Notes/Videos but in future if there would be concept of nested folders or structure then this recursive approach will seamlessly accommodate those changes in future as well . 

Below is how it looks like : 

https://github.com/code100x/cms/assets/118182376/ebb2a1a4-0349-47e3-a60d-ec0179248a13

amanbairagi30:
Closes #282 : **Dynamic Breadcrumbs**

This PR adds breadcrumbs , which will show up above the course content. These breadcrumbs will change as users move around, showing exactly where they are in the course. Making it easy to see which part of the course you're in and navigate between sections by clicking, even if the sidebar is closed, for a better user experience.

Below is how it looks like : 

https://github.com/code100x/cms/assets/118182376/14ee3656-bbca-49d0-b696-13b61cf247c0

### DIFF
--- a/src/components/BreadCrumbComponent.tsx
+++ b/src/components/BreadCrumbComponent.tsx
@@ -1,0 +1,121 @@
+'use client';
+
+import {
+  Breadcrumb,
+  BreadcrumbItem,
+  BreadcrumbLink,
+  BreadcrumbList,
+  BreadcrumbPage,
+  BreadcrumbSeparator,
+} from '@/components/ui/breadcrumb';
+import { Folder } from '@/db/course';
+import Link from 'next/link';
+import { useMemo } from 'react';
+
+export default function BreadCrumbComponent({
+  rest,
+  course,
+  fullCourseContent,
+  courseContent,
+  contentType,
+}: {
+  fullCourseContent: Folder[];
+  rest: string[];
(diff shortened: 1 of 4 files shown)


amanbairagi30:
Thank you bhaiya

PR #292, opened 2024-03-30
fixed multiple instances of singleton

### LINKED ISSUE #279: Server actions creating a new cache which is differnt from the cache being used by getPurchases
https://github.com/code100x/cms/blob/main/src/actions/refresh-db/index.ts#L76 doesnt work since a new instance of the cache class is being created (even tho we're using singletons)

mumin-khan:
### PR Fixes:
- 1 multiple instances of singleton

Resolves #279 


hkirat:
That sound super weird, testing this locally now

PR #303, opened 2024-03-31
fix(dev): improve local development experience

devsargam:
### PR Fixes:
- Improves the seed command to not fail when executed once more
- Uses `process.env.LOCAL_CMS_PROVIDER` to provide mock users in middleware and signup

image
image

### DIFF
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -1,22 +1,38 @@
 import db from '../src/db/index';
 
 async function main() {
-  await db.user.createMany({
-    data: [
-      {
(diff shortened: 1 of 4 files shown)


devsargam:
Also, @hkirat 
### Can I add a demo video which shows how to setup cms in the readme?

nimit9:
Also, @devsargam @hkirat , can we add more data in the db while seeding? [Have a gist for this](https://gist.github.com/nimit9)

devsargam:
Yeah would be really great

PR #258, opened 2024-03-22
Added highlighting for current active content on the sidebar

haddercone:
fixes #256 

This PR adds UI enhancements for the courses Sidebar component to keep track of currently active content.
It lets the user know where they are the course. 

**Video reference:**

[demo-1.webm](https://github.com/code100x/cms/assets/32171406/14865a87-1241-4ecd-9589-2556e21bd143)

PR #304, opened 2024-04-01
Fixed: Highlighting for current active content on the sidebar #256

### LINKED ISSUE #256: Highlight courses sidebar for the currently running content similar to YouTube.
Currently there is no way to know on which content we are at on the courses sidebar, this creates a bad user experience. 
It would be great , if we could implement currently running content highlighting similar to YouTube.

Screenshot from 2024-03-22 08-58-59

I'd like to work on it, if we decide to proceed with this issue.

Please let me know your thoughts on this.

Shubham-Shingala:
Sidebar appearance enhances highlighting which content is open.

Resolves #256 

- [ Y] I have performed a self-review of my code
- [ N] I assure there is no similar/duplicate pull request regarding same issue
Have the same PR #258 but in this it will auto open accordion that have content like here week 1.

https://github.com/code100x/cms/assets/75634882/24e0238c-ddf8-4bce-962d-ae856ff91a03

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -1,5 +1,5 @@
 'use client';
-import { useRouter } from 'next/navigation';
+import { useRouter, usePathname } from 'next/navigation';
 import {
   Accordion,
   AccordionContent,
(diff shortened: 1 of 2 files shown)


hkirat:
Being done in https://github.com/code100x/cms/pull/304

PR #298, opened 2024-03-30
feat: add subtitle toggle like yt

asharma991:
This Pull Request is to enable the toggle feature for subtitles
- Toggle feature for subtitles (currently for English).
Resolves #[#297] 

### Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I assure there is no similar/duplicate pull request regarding same issue

PR #322, opened 2024-04-02
fix: add exact check for language and kind, referring docs

### LINKED ISSUE #297: feature: Add toggle for captions
At this point only one thing is missing from our player, toggle for the subtitles. I understand that we have the subtitle menu and other settings in the same unordered list. But IMO it still makes sense to have a toggle (at least for English for now).

What this feature would enhance:
- Add a toggle for subtitles by pressing the 'C' key like YouTube.

asharma991:
### PR Fixes:
- Referred exact check for language and kind then only performing show/hide operation for subtitles.
- https://videojs.com/guides/text-tracks/#showing-tracks-programmatically. I could've used forEach but TextTrackList doesn't seem to have that method https://developer.mozilla.org/en-US/docs/Web/API/TextTrackList#instance_properties

Resolves #297  


asharma991:
Hey @hkirat, can we merge this?

siinghd:
closing #297 https://github.com/code100x/cms/pull/298

asharma991:
> Hi @siinghd, may I know why was this closed? This is an improvement that we need, as in #298 it was not done in the correct manner as per the docs.

@hkirat ser, this is an essential change according to the videojs docs. I don't want to iterate this again but this won't working properly for the use cases where subs are not the first item.

siinghd:
@asharma991 I closed this because we will have only one track atm and it will be english, honestly its a very small change let me think about it.

asharma991:
> @asharma991 I closed this because we will have only one track atm and it will be english, honestly its a very small change let me think about it.
> 
> also
> 
> i closed because of this #298 ... because we have only one track

Got it, but still we were not checking the kind in the first change and that's my bad. This is me cleaning up my own mess. Please think about it (:

asharma991:
Thanks for this @siinghd

PR #355, opened 2024-04-10
fix: deleting and copying of comments

### LINKED ISSUE #354: bug: The options in three dots menu of comments ,are not working
**Describe the bug**
The options in three dots menu of comments ,are not working 

**To Reproduce**
1. Go to any video 
2. Comment something 
3. Try delete and copy , both are not working 

**Expected behavior**
It should delete the comment made by the user and copy the comment text as well 

**Screenshots or GIFs**

https://github.com/code100x/cms/assets/68186313/1f0458ba-6282-40f9-8f0b-bd98d3c29b47

 - Browser Chrome Latest

amilbcahat:
### PR Fixes:
- 1 Updated the if statement in deleteCommentHandler 
- 2 Updated the handler to copy the contents of the comment

Resolves #354 
- [ Done ] I have performed a self-review of my code
- [ Done ] I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/src/actions/comment/index.ts
+++ b/src/actions/comment/index.ts
@@ -377,8 +377,8 @@ const deleteCommentHandler = async (
     }
 
     if (
-      existingComment.userId !== userId ||
-      (existingComment.userId !== userId && session.user?.role !== ROLES.ADMIN)
+      existingComment.userId !== userId &&
+      session.user?.role !== ROLES.ADMIN
     ) {
       return { error: 'Unauthorized to delete this comment.' };
     }
--- a/src/actions/comment/index.ts
+++ b/src/actions/comment/index.ts
@@ -377,8 +377,8 @@ const deleteCommentHandler = async (
     }
 
     if (
-      session.user?.role !== ROLES.ADMIN ||
-      existingComment.userId !== userId
+      existingComment.userId !== userId ||
+      (existingComment.userId !== userId && session.user?.role !== ROLES.ADMIN)
     ) {
       return { error: 'Unauthorized to delete this comment.' };
     }
--- a/src/components/Copy-to-clipbord.tsx
+++ b/src/components/Copy-to-clipbord.tsx
@@ -20,7 +20,7 @@ const CopyToClipboard = ({
 
   return (
     <div>
-      <button onClick={handleCopyClick}>
+      <button onClick={() => handleCopyClick()}>
         <div className="flex items-center gap-1">
           Copy <CopyIcon size={15} />
         </div>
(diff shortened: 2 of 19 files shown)


amilbcahat:
Hi @hkirat , can you please review this PR.

src/components/Copy-to-clipbord.tsx
```diff
@@ -20,7 +20,7 @@ const CopyToClipboard = ({
 
   return (
     <div>
-      <button onClick={handleCopyClick}>
+      <button onClick={() => handleCopyClick()}>
```
siinghd, line 23: This is not necessary as handleCopyClick has no parameters. Actually Copy component has to be modified because it does not copy the text for now

src/actions/comment/index.ts
```diff
@@ -377,8 +377,8 @@ const deleteCommentHandler = async (
     }
 
     if (
-      session.user?.role !== ROLES.ADMIN ||
-      existingComment.userId !== userId
+      existingComment.userId !== userId ||
```
siinghd, line 380: This is redundant I think,

      (existingComment.userId !== userId && session.user?.role !== ROLES.ADMIN)

should do the work

siinghd (changes requested):

src/components/Copy-to-clipbord.tsx
amilbcahat: I have modified the copy component as well it does ,copy the text now

src/actions/comment/index.ts
amilbcahat: Its not actually , this is where the issue was - 

Actually the problem was , user will always be NOT AN ADMIN , and thats why , it always showed Unauthorized on trying to delete, this AND statement checks for both -> 

1.First , user is not the creater of the comment 
2.Second , user is not an ADMIN as well !

amilbcahat:
nbb

src/components/comment/Comments.tsx
amilbcahat, line 1: Check this @siinghd  , due to this change it does copy the comment text now , working on other suggested changes as well

src/components/Copy-to-clipbord.tsx
siinghd: > I have modified the copy component as well it does ,copy the text

Let's @hkirat look at it, because at the time copy had its purpose to be like that and not copy the content, but with the introduction of roles, it should be fine to copy the content.

amilbcahat: Ok sir

siinghd:
closing this delete is working fine. Raise PR for copying

PR #363, opened 2024-04-11
added feature: Header auto hide and unhide as user scroll.

### LINKED ISSUE #361: feature:  Header auto hide and unhide as user scroll.
The header may be useless when exploring a slightly longer page or even while watching a video out of full-screen mode. 

We have seen responsive headers like on the official Medium website. The header which goes up while scrolling down and comes back on when scrolling up. 

**Additional context**

https://github.com/code100x/cms/assets/111479342/feb9e5c7-bb42-4057-9cf9-e879e5b8fead

HusainModiwala:
### PR Fixes:
- 1 added feature: Header auto hide and unhide as user scroll.

Resolves #361  

### DIFF
--- a/src/components/Appbar.tsx
+++ b/src/components/Appbar.tsx
@@ -15,7 +15,7 @@ import { ThemeToggler } from './ThemeToggler';
 import { NavigationMenu } from './landing/appbar/nav-menu';
 import SearchBar from './search/SearchBar';
 import MobileScreenSearch from './search/MobileScreenSearch';
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 export const Appbar = () => {
   const session = useSession();
--- a/src/components/Appbar.tsx
+++ b/src/components/Appbar.tsx
@@ -15,10 +15,13 @@ import { ThemeToggler } from './ThemeToggler';
 import { NavigationMenu } from './landing/appbar/nav-menu';
 import SearchBar from './search/SearchBar';
 import MobileScreenSearch from './search/MobileScreenSearch';
+import { useCallback, useEffect, useState } from 'react';
 
 export const Appbar = () => {
   const session = useSession();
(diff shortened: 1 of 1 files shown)


src/components/Appbar.tsx
```diff
@@ -15,20 +15,47 @@ import { ThemeToggler } from './ThemeToggler';
...
+  const controlAppbar = useCallback(() => {
+    if (typeof window !== 'undefined') {
+      if (window.scrollY > lastScrollY) { // if scroll down hide the navbar
+        setShow(false);
+      } else { // if scroll up show the navbar
+        setShow(true);
+      }
+
+      // remember current page location to use in the next move
+      setLastScrollY(window.scrollY);
```
hkirat, line 41: How is this not causing an infinite loop?
You're updating the `lastScrollY` here and have it as a dependency to the effect?
Is it because it's the same the second time?
Doesn't matter but we should confirm

HusainModiwala: yes you are correct it is causing an issue. Actually I had added the useCallback after I was done testing the change. And then forgot to test again. Will correct and update. Sorry for the inconvinience.

HusainModiwala:
@hkirat I have resolved the issue and now it's working fine as I have tested it. Kindly have a look. Thanks !

husainm2406:
@hkirat kindly have a look if anything else is needed.

HusainModiwala:
@hkirat can you have a look please. If all good I can merge

HusainModiwala:
@hkirat pls have a look

src/components/Appbar.tsx
hkirat: Still dont see a useCallback here?
This will cause the effect to run on every render

HusainModiwala: but using useCallback was causing infinite loop and it was not rendering that's why I removed that.

hkirat:
not needed I think, this got added in dailycode and becomes jarring if anything

PR #376, opened 2024-04-11
Fix to the responsive video player

### LINKED ISSUE #360: bug: Video player is not responsive for small screens
Screenshot 2024-04-10 183707

aakash-a-dev:
### PR Fixes: For this 
This video player is made under VideoJS file but it doesn't worth it so here the approach that I hidden the skip 15 forward & skip 15 backward button in mobile responsive mode in Portrait but on landscape mode  this buttons are visible

In Portrait Mode:

Screenshot 2024-04-12 023658

In Landscape Mode:

Screenshot 2024-04-12 023741

Resolves #360 

- [ ✔️] I have performed a self-review of my code
- [✔️ ] I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -66,6 +66,14 @@
   transform: translateX(-50%);
 }
 
+@media only screen and (max-width: 768px) and (orientation: portrait) {
+ .vjs-seek-button.skip-back.skip-15 {
+    display: none;
+  }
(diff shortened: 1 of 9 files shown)


hkirat:
good PR but had to close because of conlflicts

PR #386, opened 2024-04-12
Fix :  the type error in courseId page and extra breadcrumb separator

### LINKED ISSUE #385: Fix : Type error in the courseId page and extra breadcrumb seperator
###  **Problem Statement** 
**1st problem** : Type Error in `courses / [...coursesId] / page.tsx` , where foundContent.children was complaining for having  inappropriate type inference for `foundContent.children` , you can see in below screenshot as well

image

**Solution**  : In the imports , Content type was being imported from `@prisma/client` but it was not able to detect the children for it , but I changed the import for Content from `@prisma/client` to `@/db/course` where the Content type was already present and the error got resolved after that .

--------

**2nd Problem** : There was extra Breadcrumb Seperator in the BreadCrumbComponent.tsx  as you can see the before and after of that 

Before : When we are on the test course 1 , then too it was showing an extra Breadcrumb Seperator .

image

### **Solution**

After : Resolved it to come whenever there is any content ahead .

image

amanbairagi30:
### PR Fixes:
- 1 Type error in `courses / [...coursesId] / page.tsx` (which were marked as TODO : Fix These)
- 2 Extra BreadCrumb Separator

Before : 
image

After : 
image

Closes #385 

### DIFF
--- a/src/app/courses/[...courseId]/page.tsx
+++ b/src/app/courses/[...courseId]/page.tsx
@@ -1,12 +1,16 @@
-import { Folder, Video, getCourse, getFullCourseContent } from '@/db/course';
+import {
+  Content,
+  Folder,
+  Video,
+  getCourse,
+  getFullCourseContent,
+} from '@/db/course';
 import { getServerSession } from 'next-auth';
 import { authOptions } from '@/lib/auth';
 import { getPurchases } from '@/utiles/appx';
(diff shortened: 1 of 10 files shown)


amanbairagi30:
Hey @hkirat , Could you review this PR ?

Issue #384, opened 2024-04-12
feature: Next Video Button


Sanket-777:
**Is your feature request related to a problem? Please describe.**
Whenever i complete a video i need to go back to the track and select the next video
**Describe the solution you'd like**
When the Video gets completed there will a button next to the mark as completed button that will take me to the next video , also we can add an autoplay the next video like thing

Basant-Chahar-2003:
Can I work on this......

PR #387, opened 2024-04-12
Update VideoPlayer2.tsx

Aryam2121:
Tracks the current video index using the currentVideoIndex state variable. Accepts a list of videos as input (videoList) and uses currentVideoIndex to determine the current video. When a video ends, it checks whether there is a next video. If there is, it automatically plays the next video using handleVideoEnded. Adds a button that appears when there is a next video in the list. When clicked, it advances to the next video in the list.
//////////////
I add a button for next video for enhancement for user go to next video issue #384 
### PR Fixes:
- 1
- 2

I

### DIFF
--- a/src/components/VideoPlayer2.tsx
+++ b/src/components/VideoPlayer2.tsx
@@ -13,12 +13,18 @@ import { handleMarkAsCompleted } from '@/lib/utils';
 import { useSearchParams } from 'next/navigation';
 import './QualitySelectorControllBar';
 
-// todo correct types
+// Define the interface for video data
+interface VideoData {
+  mpdUrl: string;
+  subtitles: string;
+}
+
+// Define the props interface
 interface VideoPlayerProps {
   setQuality: React.Dispatch<React.SetStateAction<string>>;
   options: any;
   onReady?: (player: Player) => void;
-  subtitles?: string;
+  videoList: VideoData[];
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer2.tsx
```diff
@@ -29,335 +35,127 @@ const VOLUME_LEVELS: number[] = [0, 0.2, 0.4, 0.6, 0.8, 1.0];
...
-  subtitles,
   onVideoEnd,
 }) => {
   const videoRef = useRef<HTMLDivElement>(null);
   const playerRef = useRef<Player | null>(null);
   const [player, setPlayer] = useState<any>(null);
+  const [currentVideoIndex, setCurrentVideoIndex] = useState<number>(0);
   const searchParams = useSearchParams();
-  useEffect(() => {
-    const t = searchParams.get('timestamp');
```
hkirat, line 42: why this diff?

Aryam2121: The behavior and difference in this code due to searchParams depend on how the rest of your application interacts with the URL and query parameters. For example, changing the value of timestamp in the URL may affect the behavior of the video player if that value is being used for seeking to a specific time in the video. This may lead to different behavior in the component if the value of timestamp changes or is manipulated in the URL.

PR #392, opened 2024-04-12
fix: added script for video  `duration` update

### LINKED ISSUE #390: Create a simple script that takes a json with duration of videos and updates the `duration` parameter in the DB
In the `VideoMetadata` table, we have a duration field that isnt yet populated
Given a json that looks like this 
```
[{
	id: 3,
      duration: 905
}, {
	id: 4,
     duration: “”
}]
```

Where id is the `contentId` field in the `VideoMetadata` table

haddercone:
### PR Fixes: 
- 1 Added script for updating video metadata duration
- 2

Resolves #390 


package.json
```diff
@@ -86,6 +87,7 @@
     "prisma": "^5.6.0",
     "tailwindcss": "^3.3.0",
     "ts-node": "^10.9.2",
+    "tsx": "^4.7.2",
```
hkirat, line 90: Why this change?

hkirat (changes requested):
Can u remove all the lock file changes?

haddercone:
> Can u remove all the lock file changes?

on it!

PR #399, opened 2024-04-13
Fix studio command when running the project with docker.

### LINKED ISSUE #398: Running prisma studio throws error, when project is started via Docker.
**Describe the bug**
When the project is started via Docker, running the prisma studio throws error as the DB url is not running.  

**To Reproduce**
1. RUN `docker compose up` in terminal
2. RUN `prisma studio` in terminal
3. A new browser window pops-up, select any table and see the error.

**Expected behavior**
When running a command, we should be able to see all the data without any error.

**Screenshots or GIFs**
image
image

VK-RED:
### PR Fixes:
- 1 Updated the docker-compose file with 5555 port for routing on the app container
- 2 Added studio script with in the dev:docker script, so that studio runs when the container starts to run.
- 3 Added studio:docker script so that new window opens and shows the db content.

Resolves #398

### Screencast
[Screencast from 13-04-24 11:41:56 AM IST.webm](https://github.com/code100x/cms/assets/130341088/c50b233b-4797-4bb5-8ea8-e109940e0304)

-  I have performed a self-review of my code
-  I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
       - DATABASE_URL=postgresql://postgres:postgres@db:5432/cms?schema=public
     ports:
       - '3000:3000'
+      - '5555:5555'
     volumes:
       - .:/usr/src/app
       - /usr/src/app/node_modules
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "db:seed": "prisma db seed",
     "db:reset": "prisma migrate reset",
     "prepare": "husky install",
-    "studio": "prisma studio"
+    "studio": "prisma studio",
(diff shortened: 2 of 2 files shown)


hkirat:
dev:docker already has that so I think not needed

VK-RED:
No ! The dev docker has only migrate, studio and next dev. You can check here https://github.com/code100x/cms/blob/main/package.json#L17 .  Migrating doesn't populate it !!

hkirat:
Ah shoot you're right
yeah lets add that in there

VK-RED:
Yep !!

VK-RED:
Updated here #411

PR #408, opened 2024-04-14
hit checkUserEmailForPurchase in refreshDB only in prod

### LINKED ISSUE #402: bug:    TypeError.
At /src/utiles/apx-check-mail.tsThis was showing in the console

image

I have simple fix for this issue that is by adding types to the variables from which the error occured.

Thanks!

VK-RED:
### PR Fixes:

- 1. Earlier, we were hitting checkUserEmailForPurchase in refresDb in both local and prod
- 2. Since in local we don't have APPX Creds, we now only hit it in production.

Resolves #402 

### Screencast (Before fix)
[Screencast from 14-04-24 11:47:05 AM IST.webm](https://github.com/code100x/cms/assets/130341088/91966eb1-c308-4459-8cd4-2f0ff38a5cb6)

### Screencast (After fix)
[Screencast from 14-04-24 11:48:13 AM IST.webm](https://github.com/code100x/cms/assets/130341088/bc04ee74-f5ff-4973-8a42-320e08a5cd8a)

### DIFF
--- a/src/actions/refresh-db/index.ts
+++ b/src/actions/refresh-db/index.ts
@@ -24,6 +24,10 @@ export const refreshDb: RefreshDbFn = async () => {
     };
   }
 
+  if (process.env.LOCAL_CMS_PROVIDER) {
+    return { error: false, message: 'Refetched Courses' };
+  }
+
   // Only allow user to refetch every minute
   if (Cache.getInstance().get('rate-limit', [email])) {
     return {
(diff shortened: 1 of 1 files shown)


hkirat:
refreshDB should just early return for local development
No need to drill this down to the `checkUserEmailForPurchase` function
`checkUserEmailForPurchase` should only be called in prod

VK-RED:
Yup !! Agreed !!

PR #409, opened 2024-04-14
fixed typeError

### LINKED ISSUE #402: bug:    TypeError.

pratham-srivastava-07:
Resolves #402  

1. I have performed a self-review of my code

(it doesen't show status code of 500 in the console after this fix)

### DIFF
--- a/src/actions/refresh-db/index.ts
+++ b/src/actions/refresh-db/index.ts
@@ -53,7 +53,7 @@ export const refreshDb: RefreshDbFn = async () => {
       const courseId = course.appxCourseId.toString();
       const data = await checkUserEmailForPurchase(email, courseId);
 
-      if (data.data === '1') {
+      if (data) {
         responses.push(course);
       }
     });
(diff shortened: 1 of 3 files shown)


src/actions/refresh-db/index.ts
```diff
@@ -53,7 +53,7 @@ export const refreshDb: RefreshDbFn = async () => {
       const courseId = course.appxCourseId.toString();
       const data = await checkUserEmailForPurchase(email, courseId);
 
-      if (data.data === '1') {
+      if (data) {
```
hkirat, line 56: why did u update this?
This will lead to every user having access to every course

hkirat (changes requested):

src/actions/refresh-db/index.ts
pratham-srivastava-07: U r right...at first it did give access to all users but eventually it didn't after testing it multiple times...so I reduced the code to this...

PR #425, opened 2024-04-15
feature:Type safe env variables

shubham1091:
This PR creates an env.ts file for better environment variables 
we usually face the issue of the environment variable not being type-safe
to solve this issue I used t3-env to create env with correct types 
now we don't need to use '||','!' etc to remove warnings and the variables are easy to access and type-safe  


shubham1091:
@hkirat sir can you please review this PR

siinghd:
too many conflicts

PR #427, opened 2024-04-16
cleaned up the appbar

RahulMishra0722:
### PR Fixes:
- 1 Made the app bar cleaner

Previous
Screenshot 2024-04-16 at 7 20 30 AM
Now
Screenshot 2024-04-16 at 7 22 36 AM

- [ yes] I have performed a self-review of my code
- [yes ] I assure there is no similar/duplicate pull request regarding same issue


RahulMishra0722:
cleaned up the app bar check if its preffered over the current if it is not its fine

PR #436, opened 2024-04-16
feat: introducing Qs&As, reddit like tiny platform

siinghd:
### PR feat:
Introducing tiny reddit like platform where people can ask questions or post something and people can reply. 

In the prisma Schema two models are seperated (questions,answers) because with single recursive model it would be hard to get recursive answers.

Screenshot 2024-04-16 at 3 27 23 PM
Screenshot 2024-04-16 at 3 27 37 PM
Screenshot 2024-04-16 at 3 27 52 PM
Screenshot 2024-04-16 at 3 28 09 PM

- [x] I assure that the build i working in local


Neon-20:
are you thinking off adding a subreddit sort of thing as well, if yes then till which depth. @siinghd.
Something like this also engages community quite much : https://codeforces.com/blog/entry/127903

siinghd:
> are you thinking off adding a subreddit sort of thing as well, if yes then till which depth. @siinghd. Something like this also engages community quite much : https://codeforces.com/blog/entry/127903

This is just like that, looks same as that

siinghd:
@hkirat veere can you look at this?

hkirat:
wow nice
are these attached per video?
or is this a global thing?

siinghd:
@hkirat this is global thing, i think for video we could just expand the comments. We already have different type of comments.

In this somehow we could reference the video... like github does with pr's

hkirat:
Can we put this behind a globally configurable feature flag @siinghd ?
Happy to merge it but need to be able to disable it quickly if there are bugs

hkirat:
Maybe generally add a frontend feature flag functionality if possible

siinghd:
@hkirat sure let me do it

siinghd:
@hkirat  should be done, maybe its dumb but i have added env var NEXT_PUBLIC_DISABLE_FEATURES = "featurea,featureb,featurec"

if you add `qa` or `QA`, basically you disable it, from server side it will redirect the user to '/'. if you are not in the page actions also can't be used i think. let me know if this is what u were thinking

Maybe u also want to disable some things on clientside (like conditional rendering) thats why the env is with NEXT_PUBLIC

siinghd:
MMMM build has some problems with const session = await getServerSession(authOptions), let me check

siinghd:
@hkirat ok veere should have fixed the build, idk why there were issues with session types and other types but should be working now.

But do check one time https://github.com/code100x/cms/pull/436/commits/5e0429ee98c0818991f74baaf84280fe21789491

hkirat:
/bounty $600
Will be testing after merging

src/actions/commentVote/index.ts
```diff
@@ -13,12 +13,18 @@ const voteHandler = async (
   data: InputTypeHandleVote,
 ): Promise<ReturnTypeHandleVote> => {
   const session = await getServerSession(authOptions);
```
hkirat, line 15: ideally should take weather it is a vote or downvote as an input
Toggling can lead to inconsistencies if the user is in two tabs/if the tab hasn't updated

hkirat: good follow up to add @siinghd

Issue #435, opened 2024-04-16
urgent feature: Add a youtube renderer in the videoplayer


hkirat:
Right now we support
1. m3u8
2. mp4
3. DRM

videos

We need to support youtube URLs as well that play directly from Youtube

hkirat:
yup that should do it

PR #438, opened 2024-04-16
Fix: added youtube support

siinghd:
### PR Fixes:
- #435 

Added support to youtube.
It checks if the source.src is youtube url, if so it does render an Iframe containing the youtube video

- [x] I assure that the build works locally


siinghd:
This also contains PR #436

hkirat:
Diff seems huge whatsup with that

nimit9:
> Diff seems huge whatsup with that

@siinghd ser has gone crazy with this PR! This is fantastic 🔥

siinghd:
> Diff seems huge whatsup with that

I checked out from #436 thats why, for only YouTube should be just few lines. It should be : src/components/admin/ContentRendererClient.tsx only

siinghd:
@hkirat i think it can be merged safely, mostly its just the new part, it’s a tiny discussion portal

siinghd:
Closing, it's done by #461

PR #449, opened 2024-04-17
cleaned up the App-Bar- and Added Discord logo 

RahulMishra0722:
### PR Fixes:
- 1 Cleaned up the App-Bar
-  I kept the theme button separate from the toggle button, So the users can locate this feature effortlessly. Placing it prominently on the app bar ensures users are aware of its presence and can easily access it, enhancing their overall experience with the app. This approach prioritizes user convenience and aligns with my past experiences with this feature.
- l
Created a pull req Before but it didnt have logo of discord so i closed it 

Before
Screenshot 2024-04-17 at 7 26 51 AM

Now
Screenshot 2024-04-17 at 6 40 02 AM


nimit9:
Already fixed in #319

PR #457, opened 2024-04-17
Add Static Site Generation (SSG) for all pages

### LINKED ISSUE #455: feature: Static Site Generation (SSG) for all pages
Data is fetched everytime it was requested

To add SSG for all pages

mvp5464:
### PR Fixes:
- 1 Now all pages will be generated during build time
- 2 It will avoid fetching of the same data

Resolves #455

Screenshot1

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint:fix": "eslint --max-warnings 0 --config .eslintrc . --fix",
     "format:fix": "prettier --write \"**/*.{ts,tsx,json}\"",
     "format:check": "prettier --check \"**/*.{ts,tsx,json}\"",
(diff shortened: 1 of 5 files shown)


mvp5464:
@hkirat , should I add on demand revalidate, when adding new content ( so when clicking on button to add new content it will revalidate at that time (almost completed locally) .  or any other change is needed in this pr

siinghd:
@mvp5464 revalidate is needed, also solve the conflicts

mvp5464:
@siinghd , here both the page.tsx are server component and they are using searchParams, which is dynamical API, which is not compatible with generateStaticParam(). so either we need to remove searchParams or convert the page into client component. Or guide me if I'm missing something.

PR #495, opened 2024-04-27
chore: increase min char match length

### LINKED ISSUE #310: bug: Search bar exposes unwanted pages
image

Also, feels like fuzzy index is a bit high would love to hear from your side

mumin-khan:
### PR Fixes:
- 1 Improves searchresult
- 2

https://www.fusejs.io/api/options.html#minmatchcharlength
Resolves #310

### DIFF
--- a/src/app/api/search/route.ts
+++ b/src/app/api/search/route.ts
@@ -14,6 +14,7 @@ export type TSearchedVideos = {
 
 const fuzzySearch = (videos: TSearchedVideos[], searchQuery: string) => {
   const searchedVideos = new Fuse(videos, {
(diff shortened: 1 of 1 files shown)


mumin-khan:
@hkirat should work for now.

hkirat:
Dont think this fixed it
Screenshot 2024-05-13 at 2 39 54 AM

mumin-khan:
@hkirat This did what was expected. We are using fuse in backend which searches for the letters and their order and spacing between them can be different. If we want only videos starting with what we search we should use something. different. 
https://www.fusejs.io/

PR #497, opened 2024-04-27
Bug/#496

### LINKED ISSUE #496: bug: Checking off a video (to mark it watched in the side navbar) doesn't save state
**Describe the bug**
Checking off a video (to mark it watched in the side navbar) doesn't save state. If you open another tab in the side nav bar and navigate back to the page , you will find the check to be reset.

**To Reproduce**
1. Mark a video complete in the side nav-bar. 
2. Close the week tab and navigate to another week's tab and get back to the original , to find it reset. 
3. Only when you hard reload , is when the progress saved.

**Expected behavior**
The tick should maintain its state , and should not require a hard reload for the same. 

**Screenshots or GIFs**

https://github.com/code100x/cms/assets/35873312/2fc45997-c16c-4242-a1a7-898bd7a11d0c

 - Browser : Google Chrome 
 - Version : 124

manthan-sharma-23:
### PR Fixes:
- Fixes Bug #496 
- Makes a global atom (recoil state for the side-bar component and stores the check updation state)

### Approach 
- Earlier the state was only stored in micro level in the check state so when the component is unmounted the state returns to global level (previous state for checks or course).
- Implemented Recoil State for each course to manage a global state store for the side-bar component and when a video (children) is checked it will update its state in the store and rerender the component .  

Resolves #496  

- [ ✓] I have performed a self-review of my code
- [ ✓] I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -14,11 +14,9 @@ import { sidebarOpen as sidebarOpenAtom } from '@/store/atoms/sidebar';
 import { useEffect, useState } from 'react';
 import { handleMarkAsCompleted } from '@/lib/utils';
 import BookmarkButton from './bookmark/BookmarkButton';
+import { fullCourseContentAtom } from '@/store/atoms';
 
-const fullCourseContentAtom = atom({
-  key: 'full/course/content/atom/state/sidebar',
-  default: [] as Folder[],
-});
+
 
 export function Sidebar({
   courseId,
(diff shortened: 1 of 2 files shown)


siinghd:
Can't we just revalidate the page or just quick router.refresh()?

manthan-sharma-23:
Hey there , 
You mean a page reload ?

PR #500, opened 2024-04-28
Show course number for multiple courses in search results for a video

### LINKED ISSUE #499: bug: Search results redirection for same video across courses breaks
**Describe the bug**
The redirection for search results for "some" videos on https://app.100xdevs.com/ does NOT work.

Upon digging a little further on why this is happening for "some" videos and not all:
1. I found this issue appears for videos where the search API returns results that have multiple courses in them and the [UI only picks the results for the first course from the search result](https://github.com/code100x/cms/blob/main/src/components/search/VideoSearchCard.tsx#L15).
2. The Backend also returns search results across courses and NOT just the results the user has purchased courses for, which I feel is also an issue if the courses are restricted for a user.

**To Reproduce**
Steps to reproduce the behavior: The user should only have access to cohort-2 courses for this to reproducable
1. Go to https://app.100xdevs.com/
3. Search for `cookie`
4. Click on the search result item named: `16.2.1 | Auth using cookies`
5. The redirection stays on the homepage and does NOT take the user to `16.2.1 | Auth using cookies` lecture.

**Expected behavior**
The redirection of search results should work for all videos. 

**Screenshots or GIFs**
search_results_redirection_bug

 - Browser: Chrome
 - Version: 124.0.6367.91

nk4456542:
### PR Fixes:
There are a couple of fixes need here I believe (might be wrong - new here 😄):
1. A user has brought multiple courses and we need to render the course name or number to show that the video exists in two courses (Please suggest alternative solutions if you have in mind, this was the easiest I could think of to get started with)
2. A user has brought only one course but the video name is the same across courses.

The current PR state: 
- [X] Solves for 1 where the course number will be displayed as the same video name can exist across multiple courses.
- [ ] To solve for 2, search results should only contain videos the user has access to.

Resolves #499 

- [X] I assure you there is no similar/duplicate pull request regarding the same issue

### DIFF
--- a/src/components/search/SearchBar.tsx
+++ b/src/components/search/SearchBar.tsx
@@ -83,13 +83,25 @@ const SearchBar = ({ onCardClick }: { onCardClick?: () => void }) => {
     } else if (!searchedVideos || searchedVideos.length === 0) {
       return <VideoSearchInfo text="No videos found" />;
     }
-    return searchedVideos.map((video) => (
-      <VideoSearchCard
-        key={video.id}
-        video={video}
-        onCardClick={handleCardClick}
-      />
-    ));
+    return searchedVideos.map((video: TSearchedVideos) => {
+      const { id: videoId, parentId, parent, title } = video;
+      if (parentId && parent) {
+        const courses = parent.courses;
+        return courses.map((course) => {
+          const { courseId } = course;
+          return (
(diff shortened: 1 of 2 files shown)


siinghd:
@nk4456542 conflicts , can you provide images or a preview?

Issue #546, opened 2024-05-07
bug: Share on linkedin not working


hkirat:
<img width="873" alt="Screenshot 2024-05-07 at 4 42 22 PM" src="https://github.com/code100x/cms/assets/8079861/aa5185f3-2912-41ee-a822-ae3526579124">

PR #554, opened 2024-05-08
feat: generate and verify certificate on frontend

### LINKED ISSUE #545: feature: Make certificate frontend only and using react
Canvas rendering on the server is very expensive
We should just render on the client since it is faster
Can this be done? check https://github.com/abhithory/auth-certficates

Gautam-Hegde:
This PR Introduces a custom hook `useCertGen` to avoid Canvas rendering on the server(expensive ops)

## overview
certi gen

- Certificate via client side URI

uri

- Verification 

cli

Resolves #545  


Gautam-Hegde:
This PR also Fixes #546 bug

PR #560, opened 2024-05-11
Generate Certificate on Client

### LINKED ISSUE #545: feature: Make certificate frontend only and using react

VK-RED:
### PR Fixes:
- 1 Renders the certificate on browser as ( PDF & PNG )
- 2 useCert hook handles the certificate to render on client.

### Screencast
[Screencast from 11-05-24 11:48:05 AM IST.webm](https://github.com/code100x/cms/assets/130341088/8a1db27d-3a13-4c99-be24-d7ba5dd246c7)

Resolves #545 


VK-RED:
@hkirat can you check this out..

Gautam-Hegde:
Cool, I've raised a PR #554 for the same 👊

src/actions/purchases/checkCoursePurchased.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const checkCoursePurchased = async (courseId: string) => {
+  const session = await getServerSession(authOptions);
+  const user = session.user;
+
+  if (!user) {
+    return false;
+  }
+
+  const purchase = await db.userPurchases.findFirst({
```
hkirat, line 15: This wont work
Use `getPurchases` function instead
Sometimes purchases are not stored in the DB but still present in appx db

PR #572, opened 2024-05-12
added next prev feature

### LINKED ISSUE #559: feature: No next and prev button for navigating between lectures.
Right now we have to manually select lecture, for ease a Next and Prev button could be a great relief for us. Also it would be nice to highlight the current going lecture.

iamnithishraja:
### PR Fixes:
- added next video button and previous video button feature

Resolves #559 

**output**
gify

### DIFF
--- a/src/components/admin/ContentRenderer.tsx
+++ b/src/components/admin/ContentRenderer.tsx
@@ -3,6 +3,7 @@ import { getServerSession } from 'next-auth';
 import { authOptions } from '@/lib/auth';
 import { ContentRendererClient } from './ContentRendererClient';
 import { Bookmark } from '@prisma/client';
+import { getNextPrevUrls } from '@/db/course';
 
 export const getMetadata = async (contentId: number) => {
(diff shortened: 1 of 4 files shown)


src/components/admin/ContentRenderer.tsx
```diff
@@ -77,11 +79,13 @@ export const ContentRenderer = async ({
     markAsCompleted: boolean;
     bookmark: Bookmark | null;
   };
+  courseId: number;
 }) => {
   const metadata = await getMetadata(content.id);
-
+  const prevNextUrls = await getNextPrevUrls(content.id, courseId);
```
hkirat, line 85: dont do this here asynchronously. Thisll delay rendering until this is fetched, and this isnt needed for most renders

Do it in an effect maybe?

iamnithishraja: sir using cached values instead of database fetches.
please do check the last file `./db/course.ts`

PR #613, opened 2024-05-13
refactor: certificate ui & add completedAt field

### LINKED ISSUE #596: bug: revert to this type of cert
WhatsApp Image 2024-05-13 at 10 06 08

@TanmayDhobale

Gautam-Hegde:
### Context: Since we are generating certificate of type PDF ( using pdf-lib ) & PNG ( using HTMLCanvas api ) both have few standard fonts exposed for use. Either we can have a cursive font on PNG certificate but not on PDF which is inconsistent.
So in this PR I've used an overlapping web-safe font exposed by both of them i.e. Courier
#### Using a external font locally is causing loader(webpack config) issues.

- This PR also adds `completedAt` field of type Date in the schema and lastly used *legit signature* on the cert(photoshop) ;)

Resolves #596  

Screenshot (395)


Gautam-Hegde:
Awaiting a feedback. 
@hkirat

PR #634, opened 2024-05-14
bug on running the docker-compose up

Ankur1493:
There was an trailing comma, which was causing errors while running docker-compose up command, so this is just to remove it
image
Screenshot from 2024-05-14 20-58-57


victorchrollo14:
@Ankur1493  This doesn't fix the `docker-compose up` issue though

image

sasa5195 (approved):

developerTested:
issue is package node-gyp which depend on Python which does not exists hence eror

add these to docker file before running npm install
```
# Build dependencies
RUN apk add make gcc g++ python3

# install node packages
```

PR #641, opened 2024-05-15
Fix: "docker compose up: issue because of canvas package solved

MrUnknown58:
### PR Fixes:
- 1 `docker compose up` command was giving an error
trailing comma
docker issue

Resolve:
The issue was because of canvas package, it internally used some packages which wasn't correctly installed in docker. I have made changes in the Dockerfile and have added those, now it works prefectly. The minor change in the CourseCard is coz Button and ChevronRight wasn't imported in the file, have commented it for the time being.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,17 @@ WORKDIR /usr/src/app
 COPY package.json package-lock.json ./
 COPY prisma ./prisma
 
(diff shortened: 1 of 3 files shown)


hkirat:
we've gotten rid of the canvas library altogether now

PR #642, opened 2024-05-15
Questions page

Ankur1493:
Fixing the errors in questions page, things which hsingh missed, like proper pagination, adding a link in dropdown, redirecting user to login if not authenticated etc


hkirat:
Thankyou! 
Small feature and hence not assigning a bounty
Waiting for a bigger contribution from you

PR #646, opened 2024-05-16
Fixed video bookmarking unexpected behaviour (most of the video are shown bookmarked by default)

### LINKED ISSUE #635: bug: Most videos are bookmarked, without explicitly being done.
**Describe the bug**
Videos till week-12 are bookmarked by default, and errors out if bookmark being removed
**To Reproduce**
1. Uncheck bookmarks, which are previously bookmarked 
2. See error

**Expected behavior**
Expected: 
a) No previous bookmarks, which I didn't do.
b) No error if uncheck it.
**Screenshots or GIFs**

image
Works fine for new bookmarks

uttamsutariya:
### PR Fixes:
- Fixed video bookmarking unexpected behaviour (most of the video are shown bookmarked by default & removing of bookmarks isn't working perfectly)

As you can see I haven't bookmarked any video but still it's shown as bookmarked.
image

And if I try to remove that bookmark, it gives this error.
image

It was too complicated to find the bug in the first go, but eventually I found that the course content fetching doesn't have any constraints that makes sure only currently logged-in user's bookmarks are being fetched.

So I've added that constraint, that resolved this issue.

I've make sure that the issue is resolved perfectly now.

Please go through this https://github.com/code100x/cms/issues/635#issuecomment-2114674836

I've explained the issue & solution perfectly there.

Resolves #635 

- [x] I assure there is no similar/duplicate pull request regarding the same issue

### DIFF
--- a/src/db/course.ts
+++ b/src/db/course.ts
@@ -174,6 +174,10 @@ export const getNextVideo = async (currentVideoId: number) => {
 };
 
 async function getAllContent() {
+  const session = await getServerSession(authOptions);
+  if (!session?.user) {
+    return [];
+  }
   const value = Cache.getInstance().get('getAllContent', []);
   if (value) {
     return value;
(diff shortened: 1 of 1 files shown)


hkirat:
fixed already

Issue #654, opened 2024-05-17
bug: Mark as Completed is not working


dev13-suthar:
When i click on Mark as Completed it not marking my video completion..

error in network tab: error: 
"[\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"object\",\n    \"received\": \"promise\",\n    \"path\": [],\n    \"message\": \"Expected object, received promise\"\n  }\n]"

![Screenshot 2024-05-17 100635](https://github.com/code100x/cms/assets/83634894/dfe75d9b-3541-4330-b866-facce5e3de28)
[](url)

safdarkb:
still not working

PR #655, opened 2024-05-17
Fixed video mark as complete bug #654

RitikBora:
This PR is raised in reponse to the **issue #654** 

Here is what all has changed
1) In the backend route handler for the _/markAsCompleted_ route  we were parsing incorrect inputs to the zod safe parse hence resulting in a status 400. 

We need to await the response and check for the inputs.

I'm attaching the screenshot of the network logs post fix 

reference

Apart from the original issue some minor changes
1) Package.json had a minor error
2) Coursecard.tsx had some minor import issue


hkirat:
Already fixed in master

PR #663, opened 2024-05-17
geeks for geeks of 100x devs introducing 100x blogs

iamnithishraja:
### PR Fixes:
- every ed tech busiiness at some point will start their blogs i think this is the time for that.

### **lets introduce 100x blogs**
s-1

### **get notion like editing experience**
s-3

## **preview the blog**
s-4

## **post it**
s-5

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,8 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  reactStrictMode: true,
+  reactStrictMode: false,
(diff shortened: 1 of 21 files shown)


hkirat:
Hey! 
While I love this, this is already happening on https://github.com/code100x/daily-code/pulls
Pretty much is very similar to what you've built here
It'll be very hard to keep both in sync so will have to close it unfortunately

Issue #690, opened 2024-05-22
bug: dont see login button in incognito


hkirat:
<img width="1624" alt="Screenshot 2024-05-22 at 8 54 04 AM" src="https://github.com/code100x/cms/assets/8079861/ee27a6c3-8b9f-4823-8979-6e3d8df6537e">

Zafeeruddin:
![Screenshot 2024-05-22 110454](https://github.com/code100x/cms/assets/121102547/78126caa-b35e-47c5-8d35-29b07f785d2c)
Works fine for me in all browsers
This is a window machine

mvp5464:
### Sir, seems working in ubuntu machine

![Screenshot](https://github.com/code100x/cms/assets/28997328/f033858a-f730-4570-ae68-f179e983e548)

BishtAmi:
works for me in windows 
![100x](https://github.com/code100x/cms/assets/99594069/550cda68-499b-466f-896f-bcda4e784ff7)

MrUnknown58:
Hi, @hkirat, Got the issue...
While opening in a incognito tab, there's a slight delay while loading session, and we aren't showing anything to the user during that time..
![image](https://github.com/code100x/cms/assets/91674120/6beb44d8-7378-448a-b80a-4fb2cda06315)

I am picking this up and fixing this...
Cheers :)

hkirat:
Unable to reproduce

PR #691, opened 2024-05-22
Fix: 690 - Login button space on AppBar is blank for some seconds when session is loading

MrUnknown58:
### PR Fixes:
- 1 Have fixed the bug in which "Join Now" buttons on AppBar was hidden when a user opens the site in a fresh window or incognito window
- 2 Have added a placeholder loader while the session is loading, which improves the user experience
- 3 Have also improved  the code by removing isLoading used redundantly within ternary operator

Resolves Issue #690 by @hkirat 


hkirat:
THe issue was that I couldn't at all see the buttons
Cant reproduce now though

Loader isn't needed during that time

PR #698, opened 2024-05-23
Fixed Video player on hover

### LINKED ISSUE #480: feature: Video Autoplay on Hover
I would like to implement a feature that enables videos to start playing automatically when the user hovers their cursor over the video thumbnail or player. This behavior should mimic the autoplay functionality found on popular platforms like YouTube, providing a more engaging and interactive experience for users. Additionally, if the user wants to see a preview of the video before committing to playback, a short clip or teaser could be displayed when hovering over the thumbnail, allowing them to confirm the content they wish to watch.
 
** Preview **

https://github.com/code100x/cms/assets/134638667/6794422f-c1e7-41bf-b350-6ecdaf1ec6ec

Dishant1804:
### PR Fixes:
- 1 Added a responsive and dragable seek bar on hover while playing the video on hover.
- 2 Added the timeout so that the video loads on hover only when mouse is on the videothumbnail for a while(700 milliseconds).
- 3 Supports various file formats

### Description
-This PR uses **Video.Js** library which **is already being used** to play the video in main videoPlayer, i have tweaked it such a way that **only seekBar is visible along with the time elapsed**, and it is responsive so that the user can skim through it to check what the content is.
-I have also **added  a timeout to the hover** so that the video takes a very small fraction of time to load when the mouse is over it. This is done because in the current app.100xdevs.com site when we try to navigate the mouse pointer the hover triggers immediately resulting in too many calls.

Resolves  #480  
The issue **480 was reopened** to support various video formats, and this issue was linked to PR 651 which was closed because the code was very dependent on the new external library that was not used before.

I have fixed the above issue in this PR. Here is the demo video of the fix

https://github.com/code100x/cms/assets/115367435/f8c343dc-e1e9-4992-8db4-dd5dc0653c52

### DIFF
--- a/src/components/videothumbnail.tsx
+++ b/src/components/videothumbnail.tsx
@@ -1,6 +1,8 @@
-import React, { useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import VideoPreview from '@/actions/videopreview/videoPreview';
-import { useEffect } from 'react';
+import videojs from 'video.js';
+import 'video.js/dist/video-js.css';
+import Player from 'video.js/dist/types/player';
 
 const VideoThumbnail = ({
   imageUrl,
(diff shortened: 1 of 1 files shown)


Dishant1804:
@hkirat do checkout this fix, this is my very first fix in the cms repo :)
happy to get feedback and improvise the code if need be.

Dishant1804:
@hkirat do checkout this fix, this is my first PR to cms repo :)
Happy to get feedback and make changes/improvise the code if need be

Issue #570, opened 2024-05-12
feature: Navigate to search bar with keyboard


Zafeeruddin:
**Is your feature request related to a problem? Please describe.**
Its time consuming to use mouse to search videos

**Describe the solution you'd like**
A better way if search bar could be navigated with the keys.

Pritam12F:
I am working on this

PR #706, opened 2024-05-25
fix:#705 improve search bar

### LINKED ISSUE #705: feature: Improve search functionality
Feature:
1. The search bar only gives videos as search results, it would be nice if people could search the notes as well from the search bar

2.  Using the mouse to go to the search bar and manually navigating is tedious. A simple keyboard navigation could be added that lets the user search and navigate the results using keyboard only.

Solution:

I would like to modify the logic of search api route so that it also gives the notes as output. I have also made the searchbar navigable using only the keyboard

Pritam12F:
### PR Fixes:

- 1 #705
- 2 #570 

Resolves #705 

I have modified the search bar functionality so that users can also search for notes now. I have also made it navigable with just the keyboard.

Before/After:

https://github.com/code100x/cms/assets/105589734/276a3c7b-4d7f-4402-ab2a-13f797ef6271


Pritam12F:
@hkirat Please review the code sir

PR #733, opened 2024-05-29
my first commit

Jitendarmahara:
### PR Fixes:
i just removed the the bg-transparent form the style property it worked

Resolves #[722]


Jitendarmahara:
got stuck in finding the file where i need to make the changes it tookk above 1 hour this was my first commit

PR #747, opened 2024-06-01
added migration status to video metadata

### LINKED ISSUE #745: feature: Script needed to migrate all videos to bunny
We want to copy over all video data into our own systems
This would involve multiple workers running, finding a video url (720p, 360p, 240p) and copying over all the files from our current video provider to our own systems

We want to copy over files to bunny cdn - https://bunny.net/cdn/

Steps to follow
1. A script should start and pick up a video from the database, mark it `in_progres_migration` and a `migration_pickup_time`
2. It should copy over the 1080p, 720p and 360p video contents over to bunny. Video can be of type mp4 or m3u8 so we need to make sure we are actually copying over all the chunks rather than just copying over root level m3u8 file. Make sure the path is exactly the same in bunny as it was in our original video provider
3. Once everything has been copied, it should set `migrated` to `true`. 

This script will run in an autoscaling group where we will start 20 servers that should migrate all of our videos over to our own systems

Stack needed - Simple Node.js script with bunny client side APIs to upload to bunny. You will also have to write the logic to parse an m3u8

nimit9:
### PR Fixes:
Added Migration status to db for video migration

Resolves #745 

Complete Code for migration here: https://github.com/nimit9/code100x-videos

### DIFF
--- a/prisma/migrations/20240601123949_added_migration_status/migration.sql
+++ b/prisma/migrations/20240601123949_added_migration_status/migration.sql
@@ -1,0 +1,6 @@
+-- CreateEnum
+CREATE TYPE "MigrationStatus" AS ENUM ('NOT_MIGRATED', 'IN_PROGRESS', 'MIGRATED', 'MIGRATION_ERROR');
+
+-- AlterTable
+ALTER TABLE "VideoMetadata" ADD COLUMN     "migration_pickup_time" TIMESTAMP(3),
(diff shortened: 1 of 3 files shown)


hkirat:
/bounty $150
Rest of the code - https://github.com/nimit9/code100x-videos

nimit9:
Thanks @hkirat ❤️

PR #768, opened 2024-06-05
feature: added sorting of the folders

thrishank:
### PR Fixes:
- 1 Adding the Sort feature for the folders. The data is already sorted orderly. But the problem Is the latest week videos are at the last, so every time users need to scroll to last. Now you can just sort Newest First the latest added folders will come first [week 26, week 25 ... ]
- 2 And added another feature where users can sort the folders by recently watched. Suppose a user has recently watched a lecture in week 13 now they can sort by recently watched. Now The Content shows from week 13 to the last and from the beginning to 12

https://github.com/code100x/cms/assets/86158204/6122a3f7-f766-44a8-b557-32a010c9ebad

### DIFF
--- a/src/app/api/user/history/route.ts
+++ b/src/app/api/user/history/route.ts
@@ -1,0 +1,40 @@
+import { NextRequest, NextResponse } from 'next/server';
+import db from '@/db';
+
+export async function GET(req: NextRequest) {
+  const url = new URL(req.url);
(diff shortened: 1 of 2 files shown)


siinghd:
Don't think we need this, it's already sorted by weeks and you can easily scroll.
Probably there will never be so many weeks.





Closing for now.

PR #772, opened 2024-06-06
Feat/mobile auth

### LINKED ISSUE #761: feature: Move auth to be jwt only
We need to move off of next auth since we want to support react native
This would mean we use a simple `cookie` based auth for nextjs and `jwt based auth` for mobile apps.

Whenever the user signs in, we set the cookie on the domain for the browser. 

If the user logs in from mobile (/api/signin/mobile) , they get back a JWT that they can send in every future request to authenticate themselves.

nimit9:
### PR Fixes:
- 1 Added api routes for mobile login and logout
- 2 Modified `middleware.ts` to authorize mobile routes

Resolves #761 


nimit9:
@hkirat Will be testing this thoroughly, and then mark it ready for review. We're disabling auth on local, so need to modify some code to test properly

src/middleware.ts
```diff
@@ -1,23 +1,30 @@
...
-    `${process.env.NEXT_PUBLIC_BASE_URL_LOCAL}/api/user?token=${token.jwtToken}`,
-  );
 
-  const json = await user.json();
-  if (!json.user) {
-    return NextResponse.redirect(new URL('/invalidsession', req.url));
-  }
-});
+  // Mobile Auth
+  await handleMobileAuth(request);
```
siinghd, line 23: can we here get also a device type? like
deviceType:"mobile"...

so we could do
```
if(deviceType==='mobile'){
await handleMobileAuth(request);
return NextResponse.next();
}
return (nextAuthMiddleware as any)(request);
```

something like this??

siinghd (changes requested):

src/middleware.ts
nimit9: Will have to use useragent, can add it later, if needed.

Don't think we need this urgently I guess

siinghd: @nimit9 for now there will be only desktop people, and for each login request we are unnecessary calling mobileAuth  
It is also really simple to implement it, probably 2 minutes

siinghd (changes requested):

src/middleware.ts
nimit9: We'll be checking headers for user agent, and in handleMobileAuth also we're checking headers for bearer token.

But yeah will be easy to implement as you said. Happy to do it if you say so

nimit9: added the check @siinghd

siinghd: <3

src/lib/middleware-utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  return payload;
+};
+
+export const handleMobileAuth = async (request: NextRequestWithAuth) => {
+  const token = request.headers.get('Authorization')?.split(' ')[1];
+  const pathname = request.nextUrl.pathname;
+  if (token) {
+    try {
+      const payload: any = await verifyJWT(token);
+      request.nextUrl.searchParams.set('userId', payload.id);
```
siinghd, line 25: why can't be this used in logout?

src/app/api/auth/mobile/logout/route.ts
```diff
@@ -0,0 +1,28 @@
+import { NextRequest, NextResponse } from 'next/server';
+import db from '@/db';
+
+export async function POST(request: NextRequest) {
+  try {
+    const body = await request.json();
```
siinghd, line 6: i was testing this, why do i have to pass userId here? 
i saw that in https://github.com/code100x/cms/pull/772/files#diff-85f5526f057dabf2a9a7e00ec553a8049306f43a9f343ff532fb5657121c7791

u are setting userId in search params, can't we use that?

siinghd (changes requested):

src/app/api/auth/mobile/logout/route.ts
nimit9: Oh yes, we can use that, I added the url params later, and forgot to change the logout route, could you please change it? Or should I do it?

src/app/api/auth/mobile/logout/route.ts
```diff
@@ -3,26 +3,22 @@ import db from '@/db';
 
 export async function POST(request: NextRequest) {
   try {
+    // TODO: Get userid from somewhere not body
```
siinghd, line 6: lets wait for this to be solved

siinghd:
@nimit9 any update on this? would like to close and merge

PR #775, opened 2024-06-08
feat: convert server actions to api routes

### LINKED ISSUE #760: feature: Move all server actions to be api routes
We need to convert all server actions to be API routes
We also need to replace their usage throughout the codebase 
By the end, we should have 0 server actions in our codebase

devsargam:
### PR Fixes:
- Convert all the actions to routes
- [x] Courses
- [ ] Bookmark
- [ ] User

Resolves #760
Follow up for #772


devsargam:
@hkirat any other action that can be API route?

hkirat:
That is the biggest one yeah

hkirat:
Everything else can be added incrementally

PR #785, opened 2024-06-14
Fix/ certificate download Bug

### LINKED ISSUE #764: bug: Invalid Certificate being downloaded
currently If we attempt to download the Certificate before its generated, then it still downloads a png which is corrupt since the it was downloaded before generating

https://github.com/code100x/cms/assets/136328410/5f9ca8e6-b5cf-4f18-b927-a36c4cb6e685

1. go to /certificate endpoint 
2. download the certificate before it appears
(enable throttle to fast/slow 3g to reproduce if you have very good internet)

**Expected behavior**
Waits for the certificate to be generated and then downloads it

 - Browser: chrome
 - Version 125.0.6422.113 (Official Build) (arm64)

nischal-shetty2:
### PR Fixes:
- 1 Waits for Certificate to first generate and only then downloads the Certificate
- 2 displays Generating while certificate is being generated

https://github.com/code100x/cms/assets/136328410/fa0f96b8-eb5d-4763-b1c6-986c0f8b7d92

Resolves #764

### DIFF
--- a/src/components/Certificate.tsx
+++ b/src/components/Certificate.tsx
@@ -43,6 +43,8 @@ export const CertificateComponent = ({
   }, [certificateImageUrl]);
 
   const handleDownloadPNG = async () => {
+    if (loading) return;
+    setLoading(true);
     const downloadUrl = certificateImageUrl;
     console.log('downloadUrl is : ', downloadUrl);
     const a = document.createElement('a');
@@ -51,22 +53,29 @@ export const CertificateComponent = ({
     document.body.appendChild(a);
     a.click();
     document.body.removeChild(a);
+    setLoading(false);
   };
 
   const handleShareLinkedIn = async () => {
+    if (loading) return;
+    setLoading(true);
     const certificateUrl = `${window.location.origin}/certificate/verify/${certificateSlug}`;
     const postContent = `I just earned the "${course.title}" certificate on 100xDevs! Check it out: ${certificateUrl}`;
     const shareUrl = `https://www.linkedin.com/shareArticle?mini=true&url=${encodeURIComponent(
       certificateUrl,
     )}&text=${encodeURIComponent(postContent)}`;
     window.open(shareUrl);
+    setLoading(false);
   };
 
   const handleShareTwitter = () => {
+    if (loading) return;
+    setLoading(true);
     const tweetText = `I just earned the "${course.title}" certificate on 100xDevs! Check it out:`;
     const certificateUrl = `${window.location.origin}/certificate/verify/${certificateSlug}`;
     const shareUrl = `https://twitter.com/intent/tweet?text=${encodeURIComponent(tweetText)}&url=${encodeURIComponent(certificateUrl)}`;
     window.open(shareUrl);
+    setLoading(false);
   };
 
   if (certificatePdfUrl) {
--- a/src/components/Certificate.tsx
+++ b/src/components/Certificate.tsx
@@ -11,7 +11,9 @@ import { Button } from './ui/button';
 import { FaFileImage, FaLinkedin, FaTwitter } from 'react-icons/fa';
 import { useGenerateCertificate } from '@/hooks/useCertGen';
 import { OneCertificate } from '@/utiles/certificate';
-import { useMemo } from 'react'; //used to fix maximum update depth exceeded err
+import { useMemo, useState, useEffect } from 'react';
+
+type ActionType = 'downloadPNG' | 'shareLinkedIn' | 'shareTwitter';
 
 export const CertificateComponent = ({
   certificateId,
@@ -19,6 +21,10 @@ export const CertificateComponent = ({
   course,
   userName,
 }: OneCertificate & { userName: string }) => {
+  const [triggeredAction, setTriggeredAction] = useState<ActionType | null>(
+    null,
+  );
+
   const certificateDetails = useMemo(
     () => ({
       certificateId,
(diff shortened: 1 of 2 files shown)


nischal-shetty2:
@siinghd @hkirat could you review this please

src/components/Certificate.tsx
```diff
@@ -34,77 +40,80 @@ export const CertificateComponent = ({
...
-  // const handleDownloadPDF = async () => {
-  //   const downloadUrl = certificatePdfUrl;
-  //   const a = document.createElement('a');
-  //   a.href = downloadUrl!;
-  //   a.download = 'certificate.pdf';
-  //   document.body.appendChild(a);
-  //   a.click();
-  //   document.body.removeChild(a);
-  // };
+  useEffect(() => {
```
siinghd, line 43: Why this?

src/components/Certificate.tsx
```diff
@@ -11,14 +11,20 @@ import { Button } from './ui/button';
...
+
+type ActionType = 'downloadPNG' | 'shareLinkedIn' | 'shareTwitter';
 
 export const CertificateComponent = ({
   certificateId,
   certificateSlug,
   course,
   userName,
 }: OneCertificate & { userName: string }) => {
+  const [triggeredAction, setTriggeredAction] = useState<ActionType | null>(
```
siinghd, line 24: What is this?

siinghd (changes requested):
I think we dont need any of this, please just use state 

const [loading,setLoading] = useState(false);

in the functions just set it true initially and false at the end

nischal-shetty2:
> I think we dont need any of this, please just use state
> 
> const [loading,setLoading] = useState(false);
> 
> in the functions just set it true initially and false at the end

had made a pr with this exact solution but thought it would be better if the user can click download and only after the certificate is generated does the download happen. ill make the requested changes and update @siinghd

nischal-shetty2:
@siinghd @hkirat updated code with requested changes

https://github.com/code100x/cms/assets/136328410/68bf4d6c-e17a-4679-8f80-fcf4d12eeef4

src/components/Certificate.tsx
nischal-shetty2: changed in latest commit

src/components/Certificate.tsx
```diff
@@ -34,15 +36,11 @@ export const CertificateComponent = ({
...
-  //   document.body.removeChild(a);
-  // };
+  useEffect(() => {
+    if (certificateImageUrl) {
+      setLoading(false);
+    }
+  }, [certificateImageUrl]);
 
   const handleDownloadPNG = async () => {
     const downloadUrl = certificateImageUrl;
```
siinghd, line 46: we can also here add setLoading(true)

and at the end of the function set loading to false?

this way if the user clicks download multiple times the image doesnt get dowaloaded multiple times right ? 

do it for all the three functions

nischal-shetty2: but then the buttons for share on linkedin and twitter would be disabled as well right? did you mean for me to make three states for three functions seperately?

nischal-shetty2: @siinghd should i put a 1 second timeout after the click to fix this instead? or go with the 3 separate states? or did you mean some other way?

siinghd: > @siinghd should i put a 1 second timeout after the click to fix this instead? or go with the 3 separate states? or did you mean some other way?

You already have loading state

nischal-shetty2: > You already have loading state

@hkirat yes but that will disable all three of the buttons after clicking on download

nischal-shetty2: @siinghd

src/components/Certificate.tsx
```diff
@@ -53,58 +53,71 @@ export const CertificateComponent = ({
...
+    setLoading(true);
     const tweetText = `I just earned the "${course.title}" certificate on 100xDevs! Check it out:`;
     const certificateUrl = `${window.location.origin}/certificate/verify/${certificateSlug}`;
     const shareUrl = `https://twitter.com/intent/tweet?text=${encodeURIComponent(tweetText)}&url=${encodeURIComponent(certificateUrl)}`;
     window.open(shareUrl);
+    setLoading(false);
   };
 
-  console.log('Url is : ', certificatePdfUrl);
+  if (certificatePdfUrl) {
```
siinghd, line 81: remove this please

siinghd (changes requested):

src/components/Certificate.tsx
nischal-shetty2: > remove this please

sure, you meant the if check right? or the whole console.log?? @siinghd

src/components/Certificate.tsx
```diff
@@ -78,9 +78,7 @@ export const CertificateComponent = ({
     setLoading(false);
   };
 
-  if (certificatePdfUrl) {
-    console.log('Url is : ', certificatePdfUrl);
-  }
+  console.log('Url is : ', certificatePdfUrl);
```
siinghd, line 81: sir remove the console please

siinghd (changes requested):

nischal-shetty2:
@siinghd done sir! got confused because of the line you wrote the review in which was at the if check and since the console already existed i thought it was necessary

nischal-shetty2:
@siinghd done with all the changes you can merge this now

nischal-shetty2:
@siinghd waiting to be merged

nischal-shetty2:
@siinghd check it out , done with all the changes

PR #807, opened 2024-07-01
Avatar page to change the Display Picture (only frontend)

sammamama:
Added a way for users to choose their avatar using naruto characters. Created a UI for the page and soon will add the feature where the actual avatar would change according to user selection.

### PR benefits:
- 1 Allows user to choose avatar.
- 2 Looks cool

### Check it out
- Head over 'app.100xdevs/avatar' endpoint.

Preview looks something like:

https://github.com/code100x/cms/assets/167903808/4b0b7149-5696-42c6-a07e-8a9cf4c3b6cf


siinghd:
what is this?

PR #821, opened 2024-07-10
Migrate db- store date and timestamp of video

### LINKED ISSUE #820: Create a script that migrates the DB to store timestamps and dates of all videos
This is the csv that has all the data
[cohort2.json](https://github.com/user-attachments/files/16164721/cohort2.json)

Ha4sh-447:
### PR Fixes:
-  Adds date and timestamp fields to VideoMetaData table

Resolves #820 


prisma/migrations/20240710170832_add_date_timestamp_to_video_metadata/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "VideoMetadata" ADD COLUMN     "date" TIMESTAMP(3),
+ADD COLUMN     "timestamp" INTEGER;
```
heyFahad, line 3: Column `date` and `timestamp` sound very similar in this table. I think the column you wanted to add is `duration`, as per the given data.

prisma/schema.prisma
```diff
@@ -121,6 +121,8 @@ model VideoMetadata {
   migrated_video_1080p_mp4_1 String?
   migrated_video_360p_mp4_1  String?
   migrated_video_720p_mp4_1  String?
+  timestamp Int?
```
heyFahad, line 124: Same here. `timestamp` does not reflect the correct data by its name.

src/scripts/updateVideoDateTime.ts
```diff
@@ -0,0 +1,35 @@
...
+export function readJson(path: string) {
+  const data = JSON.parse(readFileSync(path, 'utf-8'));
+  return data;
+}
+
+async function updateVideoDateTime(jsonPath: string) {
+  const data = readJson(jsonPath);
+  try {
+    await db.$transaction(async (tx) => {
+      for (const { id, date, timestamp } of data) {
```
heyFahad, line 14: The data in [cohort2.json](https://github.com/code100x/cms/blob/491ab2862eb834ca1433b764cb709a44705c19ea/src/scripts/dateAndTimestampData/cohort2.json) file does not have any `timestamp` key, it is `duration`.

src/scripts/updateVideoDateTime.ts
```diff
@@ -0,0 +1,35 @@
...
+        await tx.videoMetadata.update({
+          where: {
+            contentId: id,
+          },
+          data: {
+            timestamp: timestamp as number,
```
heyFahad, line 20: `timestamp` will be undefined here since it does not exist in the JSON data.

prisma/migrations/20240710170832_add_date_timestamp_to_video_metadata/migration.sql
Ha4sh-447: `duration` column is already present in VideoMetadata so should I remove the `timestamp` column? Also I will change `date` column's data type to store just date in schema.prisma.

prisma/schema.prisma
Ha4sh-447: Okay so I made a mistake in data type.  `DateTime` will be appropriate right?

src/scripts/updateVideoDateTime.ts
Ha4sh-447: I will make changes to this.

prisma/migrations/20240710170832_add_date_timestamp_to_video_metadata/migration.sql
heyFahad: I'm not sure about that, I just cloned the repo and was trying to understand the codebase when I saw your PR. I hope someone from the frequent contributors can shed more light on this

Ha4sh-447: Okay, I will go ahead with other changes then

Ha4sh-447 (commented):
Made changes

PR #838, opened 2024-07-27
feat: added vizolv chat

auspy:


devsargam:
Can u be a bit more verbose on what are u implementating? Can we have some screenshots and more contexts. Closing for now feel free to reopen if u think it's worth it

PR #839, opened 2024-07-27
Feature/commentype

### LINKED ISSUE #238: Type of  Comment DropDown shows incorrect name on change.

Anurag12133:
### PR Fixes:
- 1 Comment TYpe to default for Intro comments
- 2 Single compnent of page

Resolves #238  

- [X
Screenshot from 2024-07-27 20-44-11
] I assure there is no similar/duplicate pull request regarding same issue


siinghd:
Sir what's with the changes?

Anurag12133:
@siinghd Actually when click on Intro comment , the bar where this text was showed as INTRO only but in case of All comments it is showing what its default. So i changed the type of Intro comment to default so the text is same as Intro comments not only INTRO

siinghd:
@Anurag12133 why so many changes? can you fix that?

Anurag12133:
@siinghd yes sure

PR #846, opened 2024-08-03
Fixed modifier action propogation

SujithThirumalaisamy:
## Expected behaviour:
- The default click alt + number should only change the browser tabs. 
## Previous behaviour:
- Both the tab switch and the videoplayer seek to the respective number percentage was happening
## To reproduce:
- Try to play a video in app.100xdevs.com
- Try clicking alt + n ( where n is the index of a tab in your browser )
- This will automatically seek the video to the (n*10)%
- Which should only happen when n pressed alone without any modifier keys


SujithThirumalaisamy:
Git diff looks crazy. Just cleaned up the code a bit

siinghd:
Thank you sir!

PR #852, opened 2024-08-05
Removed the scroll wheel from the sidebar but the sidebar is still scrollable

iamlikeaproforyou:
Removed the scroll wheel from the side bar

### PR Fixes:
- 1 Removes the sidebar scroll wheel

Here is a demonstrated video for the same :
https://github.com/user-attachments/assets/78946f80-4ef1-4abd-9917-3c2a155c9c27
nevermind the duplicate week 1 columns i have did it for testing purposes and removed it 

i would like to suggest to remove the footer from this page or redesign the footer so that it does not come below the sidebar similiar to other course websites like udemy

- [ yes] I have performed a self-review of my code
- [yes ] I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,7 +94,7 @@
         "postcss": "^8",
         "prettier": "^3.2.4",
         "prettier-plugin-tailwindcss": "^0.6.1",
-        "prisma": "^5.13.0",
+        "prisma": "^5.17.0",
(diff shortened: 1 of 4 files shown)


siinghd:
@iamlikeaproforyou why so many changes?

PR #905, opened 2024-08-18
Fix #904: Infinite Requests to Certificate when viewing it

### LINKED ISSUE #904: bug: Infinite Requests goes to Certificate when viewing it
**Describe the bug**
Infinite Requests is being sent to certificate

**To Reproduce**
1. go to any certificate link here's one https://app.100xdevs.com/certificate/verify/UXZKOQCC 
Take your development skills from 0 to 100 and join the 100xdevs community - Google Chrome 2024-08-18 16-39-08

xKrishnaSaxena:
### PR Fixes:
- Resolves #904  Infinite Requests to Certificate when viewing it

### Checklist before requesting a review:
- [x] I have performed a self-review of my code.
- [x] I assure there is no similar/duplicate pull request regarding the same issue.

### Summary:
- **Issue**: Infinite requests are being sent to the certificate endpoint when viewing a certificate.
- **Fix**: Adjusted the `useEffect` in `useCertGen.ts` to prevent the infinite loop by adding the correct dependencies and optimizing the fetch logic.

### Testing:
- I tested the fix locally and verified that the infinite requests issue is resolved.
https://github.com/user-attachments/assets/c4696e3f-0821-43bd-abd7-b3ad49bcb266

### DIFF
--- a/src/hooks/useCertGen.ts
+++ b/src/hooks/useCertGen.ts
@@ -26,6 +26,7 @@ export function useGenerateCertificate({
   const [downloadDate, setDownloadDate] = useState<Date | null>(null);
 
   useEffect(() => {
+    if (!certificateDetails) return;
     async function generateCertificateAndImage() {
       setGenerating(true);
 
@@ -41,7 +42,7 @@ export function useGenerateCertificate({
     }
 
     generateCertificateAndImage();
-  }, [certificateDetails]);
+  }, []);
 
   async function generateCertificate() {
     try {


src/hooks/useCertGen.ts
```diff
@@ -41,7 +42,7 @@ export function useGenerateCertificate({
     }
 
     generateCertificateAndImage();
-  }, [certificateDetails]);
+  }, []);
```
siinghd, line 45: but now this will not re-run if certificateDetails are available right?

xKrishnaSaxena: But once we get the certificate details on page reload using useEffect then , i think it should not run useGenerateCertificate , we can use a flag instead of this but then we will need to add a button that will regenerate if the details are not availble , i can add that if you want me to and then the dependency array will have certificateDetails.

PR #967, opened 2024-08-21
added a feature like youtube on which on hovering over videobar you c…

harikeshpandey:
created a video preview like youtube on hovering over video bar


siinghd:
@harikeshpandey can you show any video about it? 

Otherwise will test it later

siinghd:
@harikeshpandey also fix lock file please

Issue #931, opened 2024-08-20
bug: all date data show in one date 10august2024  @kirat_tw   and  @100xDevs   team


moteeullahazmi:
![image](https://github.com/user-attachments/assets/a8030f98-e686-46e5-81b6-5b676ab38100)

pruthviraj7714:
@hkirat I will work on this

Allan2000-Git:
What's the bug actually here? @moteeullahazmi 

Maybe it was actually added on the same day

avayyyyyyy:
working on it....

avayyyyyyy:
@siinghd sir can you please look into my pr #975 😇

Thank you!

PR #975, opened 2024-08-21
#931 Fixed Date Formatting Issue for Course Content Card from hardcoded da…

avayyyyyyy:
#931 bug: all date data show in one date 10august2024 #931

**Issue**:
All content items were showing the same date ("10 August 2024").
The date wasn't being correctly pulled from the createdAt field in the database.

Screenshot 2024-08-21 at 10 20 20 PM

**Solution**:
Updated the code to use the correct createdAt date from the database.
Now, the date is displayed as "Aug 21 2024," or whatever the actual creation date is for each content item.

Screenshot 2024-08-21 at 10 20 45 PM

**Changes Made**:
Passed the createdAt date to the CourseView component.
Used toLocaleDateString() to format the date in the "day month year" style.
Added a fallback in case createdAt is missing.

Thank you!


src/components/ContentCard.tsx
```diff
@@ -90,7 +92,14 @@ export const ContentCard = ({
         <div className="space-y-2">
           <h3 className="text-bold text-lg tracking-normal">{title}</h3>
           <h4 className="text-bold text-sm tracking-normal text-[#64748B]">
-            Posted on: 10 Aug 2024
+            Posted on:{' '}
+            {postedDate
+              ? new Date(postedDate).toLocaleDateString('en-US', {
```
siinghd, line 97: can we use dayjs? we already have it

avayyyyyyy: Sure Adding that...🫡

avayyyyyyy:
@hkirat

Issue #1016, opened 2024-08-25
bug: Add back a `refresh db` button


hkirat:
it was present in the old design 
<img width="618" alt="Screenshot 2024-08-25 at 1 09 43 PM" src="https://github.com/user-attachments/assets/5ca07329-3a86-4d19-96a6-8fb0f944fbdf">

hkirat:
Put it right above the logout button

rameshguditalu:
@hkirat What does refresh db button should do onclicking of the button?

toufiqfarhan0:
@hkirat @siinghd can you please review it ?

rameshguditalu:
@hkirat @siinghd Kindly review this PR - https://github.com/code100x/cms/pull/1019

hkirat:
Closing since the diff is so big, please dont merge lock files
Merged https://github.com/code100x/cms/pull/1019 instead

PR #1019, opened 2024-08-25
Added RefreshDB button

rameshguditalu:
Added Refresh Button 
- Fixes issue #1016


nimit9:
The ReefreshDB component has extra things which might not be needed

PR #1078, opened 2024-08-29
chore: only allow `pnpm` as the package manager for the project

devsargam:
### PR Fixes:
- Only allow pnpm as the package manager in the project
- Remove other unwanted lock files: `package-lock.json` & `yarn.lock` 

#### When using npm

There's a bug in npm cli which runs preinstall hook after the installation so we don't get a good error message like when using yarn but gets the job done

ref: https://github.com/npm/cli/issues/2660#issuecomment-794363899

cc @hkirat @siinghd 


siinghd:
@devsargam thank yo usir

PR #1119, opened 2024-09-02
fix(#1115)/authenticated users access for routes

### LINKED ISSUE #1115: bug: Unauthenticated users can access the /questions page
**Describe the bug**
When a user is not logged in, they are still able to access the /questions page. This is not the expected behavior.

**To Reproduce**
1. logout youself if you are logged in at https://app.100xdevs.com
2. go to the page /questions

**Expected behavior**
When a user is not logged in and attempts to access the /questions page, they should be redirected to the sign-in page at https://app.100xdevs.com/signin.

 - Browser firefox
 - Version 129.0.2 (64-bit)

toufiqfarhan0:
### PR Fixes:
- Authenticated user access for these( "/question", "/questions", "/watch-history" ) routes .

Resolves #1115 

https://github.com/user-attachments/assets/3cbe5a36-1b44-4562-a52b-7e165292c23c


toufiqfarhan0:
@hkirat @siinghd  we have two different routes "/question" and "/questions". We can just authenticate the "/question" routes and not "/questions" this and use it on the main page like this: 

what do you think? 

https://github.com/user-attachments/assets/4b36f441-aed1-47a5-8d5b-6dc370e813d8

maheshsp1809:
good one @toufiqfarhan0 I too solved the issue but you solved for all other routes too// one question I had is did u also get tsconfig eslint something when u switched to pnpm ?

KIWILmao:
To create a question, you need to be logged in. Therefore, it’s better not to display the 'Question' button on the landing page for users who are not logged in. Hiding the create question button on /questions route in this case would be more appropriate.

toufiqfarhan0:
> good one @toufiqfarhan0 I too solved the issue but you solved for all other routes too// one question I had is did u also get tsconfig eslint something when u switched to pnpm ?

yesss I don't why that also commiting but not showing what is that change in it.

toufiqfarhan0:
> To create a question, you need to be logged in. Therefore, it’s better not to display the 'Question' button on the landing page for users who are not logged in. Hiding the create question button on /questions route in this case would be more appropriate.

Yeah you're right what if some random guy who don't know what is (https://app.100xdevs.com/) like he/she wants to know and ask questions like how to buy course ?

KIWILmao:
It might be worth considering creating a separate component for the scenario you have mentioned, like an 'FAQ' section or something similar. Also, I've noticed there are two bookmark routes as well.

toufiqfarhan0:
> It might be worth considering creating a separate component for the scenario you have mentioned, like an 'FAQ' section or something similar. Also, I've noticed there are two bookmark routes as well.

yeahh but we can access it without loggin in

KIWILmao:
for me it's keep loading and says 504 Gateway Time-out

KIWILmao:
@toufiqfarhan0 i checked your solution you should have just added routes in ```middleware.ts/matcher``` Is there a specific reason you chose to check the session on every page instead?

toufiqfarhan0:
> @toufiqfarhan0 i checked your solution you should have just added routes in `middleware.ts/matcher` Is there a specific reason you chose to check the session on every page instead?

I too got confused in that but @hkirat he help with that!

Unfortunately in Next.js, middleware.ts is used to execute server-side logic before requests reach your application's routes. It allows you to modify the request or response, redirect, or even block the request altogether. However, middleware.ts does not run on the client side and so, cannot directly use React components like <Redirect />(clientside)

KIWILmao:
> > @toufiqfarhan0 i checked your solution you should have just added routes in `middleware.ts/matcher` Is there a specific reason you chose to check the session on every page instead?
> 
> I too got confused in that but @hkirat he help with that!
> 
> Unfortunately in Next.js, middleware.ts is used to execute server-side logic before requests reach your application's routes. It allows you to modify the request or response, redirect, or even block the request altogether. However, middleware.ts does not run on the client side and so, cannot directly use React components like (clientside)



I just added a bunch of paths in matcher, and it seems to be working.
```
export const config = {
  matcher: [
    '/courses/:path*',
    '/question/:path*',
    '/questions/:path*',
    '/bookmark',
    '/bookmarks',
    '/watch-history',
  ],
};
```

toufiqfarhan0:
> > > @toufiqfarhan0 i checked your solution you should have just added routes in `middleware.ts/matcher` Is there a specific reason you chose to check the session on every page instead?
> > 
> > 
> > I too got confused in that but @hkirat he help with that!
> > Unfortunately in Next.js, middleware.ts is used to execute server-side logic before requests reach your application's routes. It allows you to modify the request or response, redirect, or even block the request altogether. However, middleware.ts does not run on the client side and so, cannot directly use React components like (clientside)
> 
> I just added a bunch of paths in matcher, and it seems to be working.
> 
> ```
> export const config = {
>   matcher: [
>     '/courses/:path*',
>     '/question/:path*',
>     '/questions/:path*',
>     '/bookmark',
>     '/bookmarks',
>     '/watch-history',
>   ],
> };
> ```

Yeaah! Raise a PR 🤝

Issue #1156, opened 2024-09-07
bug: Store ips of the user in the db along with token, make sure token is valid for that ip


hkirat:
If someone steals my jwt, they should not be able to access my dashboard
we should check if the ip is the same as the one they logged in with

ItsFlash10:
on it

farjad-akhan:
On it! #trying

MeerUzairWasHere:
@hkirat We have to add model for IP, where we save all ips or just one more column in user table?
and we can stop him to login if ip are not same?

NawinKumarSharma:
@hkirat , do we need to save ipAddress in session or userschema?

devsargam:
@hkirat do we want this? really

rishav-goswami:
> If someone steals my jwt, they should not be able to access my dashboard we should check if the ip is the same as the one they logged in with

Hey @hkirat and @devsargam ,

Instead of IP tracking we should use `http-only` cookies for `JWT `storing and its combination with `CORS`. Which can mitigate the mentioned issues of stealing `JWT` and accessing content. 

pov:- _In case if `JWT `is stolen even with this implementation then attacker is not able to access dashboard because he can't set the cookie to our origin, ultimately `CORS` will effect if he tries from other origin_

I can work on this If its resolves what you mentioned.

PR #1158, opened 2024-09-07
fix: JWT IP bound  for enhanced security

ymohit1603:
### PR Fixes:
#1156 


MeerUzairWasHere:
I think, he said, if someone steals my jwt. but you're passing ip in jwt also? or I'm missing anything?

ymohit1603:
Yeah,int this case we are signing jwt with ip address along with the traditional payload thing in the middleware when we are verifying the signed jwt it will return us the payload and the previous ip address used to sign it and then we can check if current ip is equal to the ip that signed the jwt.I think there is no need to store ip address in db when we can just sign the token with ip itself.

ymohit1603:
Although that person knows jwt but he doesn't know the secret used to sign and he will be unable to decode the jwt to get the ip

MeerUzairWasHere:
Makes sense.

Smnthjm08:
Question: Wouldn't it be better if we store token and IP Address in redis? @ymohit1603

ymohit1603:
I am a bit confused why we need to store the JWT tokens in db. JWT is used for stateless authentication where we don't need to store the tokens in db or anywhere else , but instead we let clients keep it and we just need to verify the signature on server.
Although i think redis is a more efficient way for storing jwt than relational db.

PR #1162, opened 2024-09-08
feat: saved user IP and added a check to verify it

### LINKED ISSUE #1156: bug: Store ips of the user in the db along with token, make sure token is valid for that ip

ItsFlash10:
Resolves #1156

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -17,11 +17,11 @@ export default withAuth(async (req) => {
     return NextResponse.redirect(new URL('/invalidsession', req.url));
   }
   const user = await fetch(
-    `${process.env.NEXT_PUBLIC_BASE_URL_LOCAL}/api/user?token=${token.jwtToken}`,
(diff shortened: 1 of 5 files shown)


src/middleware.ts
```diff
@@ -8,6 +8,11 @@ export const config = {
...
   if (process.env.LOCAL_CMS_PROVIDER) return;
   const token = req.nextauth.token;
+  const userIp =
+    req.headers.get('x-forwarded-for') ||
+    req.ip ||
+    req.headers.get('x-real-ip');
+
   if (!token) {
     return NextResponse.redirect(new URL('/invalidsession', req.url));
   }
```
hkirat, line 18: In this endpoint itself, send the ip and do the check there
Dont do it saparately below


Or


move all the logic for token check here as well

Dont have two saparate places for doing the check

ItsFlash10: @hkirat done

PR #1172, opened 2024-09-09
fix:bookmark

paahaad:
### PR Fixes:
- 1 fix bookmark security issue
[Screencast from 10-09-24 01:27:44 AM IST.webm](https://github.com/user-attachments/assets/63c27551-28db-48b6-aa7a-f9901688ddad)

### DIFF
--- a/src/app/(main)/(pages)/layout.tsx
+++ b/src/app/(main)/(pages)/layout.tsx
@@ -1,9 +1,16 @@
 import React from 'react';
+import { getServerSession } from 'next-auth';
+import { Redirect } from '@/components/Redirect';
 
 interface Props {
   children: React.ReactNode;
 }
 
-export default function MainLayout(props: Props) {
+export default async function MainLayout(props: Props) {
+  const session = await getServerSession();
+
+  if (!session?.user) {
+    return <Redirect to={'/'} />;
+  }
   return <div className="w-full py-16">{props.children}</div>;
 }
--- a/src/db/bookmark.ts
+++ b/src/db/bookmark.ts
@@ -8,6 +8,9 @@ export const getBookmarkDataWithContent = async (): Promise<
 > => {
   const session = await getServerSession(authOptions);
   const userId = session?.user.id;
+  if (!userId) {
+    return { error: 'Unauthorized' };
+  }
 
   return await db.bookmark.findMany({
     where: {
--- a/src/components/CourseCard.tsx
+++ b/src/components/CourseCard.tsx
@@ -13,10 +13,7 @@ export const CourseCard = ({
   onClick: () => void;
 }) => {
   const router = useRouter();
-  const imageUrl =
-    course.imageUrl
-      ? course.imageUrl
-      : 'banner_placeholder.png';
+  const imageUrl = course.imageUrl ? course.imageUrl : 'banner_placeholder.png';
   return (
     <div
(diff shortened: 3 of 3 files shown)


paahaad:
@hkirat This requires your immediate attention.

ItsFlash10:
It's good to have a check there but it could only happen if you already have an account and you have logged in previously, tried in incognito it didn't work. Prolly some caching issue not sure.

src/app/(main)/(pages)/layout.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { getServerSession } from 'next-auth';
+import { Redirect } from '@/components/Redirect';
 
 interface Props {
   children: React.ReactNode;
 }
 
-export default function MainLayout(props: Props) {
+export default async function MainLayout(props: Props) {
+  const session = await getServerSession();
```
ItsFlash10, line 10: this check will be added at /app/(main)/layout.tsx

It works here as well but you can see the screen for a second and then it redirects, but if you add this check in the layout above, it'll directly redirect without showing the screen for a split second.

src/db/bookmark.ts
```diff
@@ -8,6 +8,9 @@ export const getBookmarkDataWithContent = async (): Promise<
 > => {
   const session = await getServerSession(authOptions);
   const userId = session?.user.id;
+  if (!userId) {
```
ItsFlash10, line 11: this isn't required as the layout will navigate the user before even this function gets invoked

paahaad: @ItsFlash10  server function are http under the hood. its good pratice to make sure that api is also protected.

src/app/(main)/(pages)/layout.tsx
paahaad: @ItsFlash10 it is the same. since page is a route group. it will not matter plus my thought was to project the pages only not everything inside the main folder.

paahaad:
> It's good to have a check there but it could only happen if you already have an account and you have logged in previously, tried in incognito it didn't work. Prolly some caching issue not sure.

I have tested it. checked the code. This issue persist in incognito. I'm able to see every offline vidoes that @hkirat has uploaded without login.

src/app/(main)/(pages)/layout.tsx
ItsFlash10: both the layouts serve the same set of pages and yes it works as I mentioned but the issue is the redirections delay but its fine here as well

ItsFlash10:
> > It's good to have a check there but it could only happen if you already have an account and you have logged in previously, tried in incognito it didn't work. Prolly some caching issue not sure.
> 
> I have tested it. checked the code. This issue persist in incognito. I'm able to see every offline vidoes that @hkirat has uploaded without login.

True, it worked this time, I was navigating from the URL not the icon prolly that's why. Really good catch dude.

devsargam:
Damn

src/db/bookmark.ts
devsargam: Yes I agree with @paahaad here, server actions will get converted into sort of `grpc` like calls

devsargam:
@paahaad can't thank u enough sir. I will let @hkirat know about this. Meanwhile I have deployed the fix. LMK if u find anything in this deployment too.

PR #1199, opened 2024-09-11
fix: thumbnail-server-action

### LINKED ISSUE #1198: bug: Thumbnail component leaks video metadata
**Describe the bug**
Thumbnail component exposes video content to users who haven't purchased the course.

paahaad:
### PR Fixes:
- 1 Course page will not leak the content
- 2 Stop unwanted multiple db calls made via server functions.

Resolves #1198

Context: We often think that the server function uses some sort of RPC call under the hood, but that's not the case. Next.js uses an HTTP call to serialize the data to the server. You can see this request(calling server function) in the network tab; the URL will be the same as the current path along with the header 'next-action': "<$7e6r763s34534f345w3434>". This key is associated with the server function, and you can find the function association in the .next folder.

Issue: There was feature which play the video content in the thumbnail which uses server function to get the video Metadata. that server function was unprotected. all I needed was a session with the server and I can get the any course video content. for example I have bought the 1-100 cohort.I was able to see the videos of cohort-3 and cohort-1.

Fix: since we are not using that feature(video player on the thumbnail). I have commented out the server function. may be we might implement that in future with proper authentication.

https://github.com/user-attachments/assets/606855ab-5cde-4b95-a8ae-0776a7d2f486

### DIFF
--- a/src/actions/videopreview/videoPreview.tsx
+++ b/src/actions/videopreview/videoPreview.tsx
@@ -1,18 +1,19 @@
 'use server';
-import db from '@/db';
+// import db from '@/db';
 
 export default async function VideoPreview({
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   contentId,
 }: {
   contentId: number;
 }) {
-  const videoMetadata = await db.videoMetadata.findFirst({
-    where: { contentId },
-    select: { video_360p_1: true },
-  });
+  // const videoMetadata = await db.videoMetadata.findFirst({
+  //   where: { contentId },
+  //   select: { video_360p_1: true },
+  // });
 
-  if (videoMetadata) {
-    return videoMetadata.video_360p_1;
-  }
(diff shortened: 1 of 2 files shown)


ItsFlash10:
Bro's on a mission now xD
cc: @devsargam @hkirat this needs triage

paahaad:
Hi @hkirat & @devsargam This need need your attention.

devsargam:
Checking this one sir.
cc @siinghd

siinghd:
@paahaad hi thanks for point out this, after commenting out does the thumbnails and everything as before?

paahaad:
> and

Yes. I have tested. it will not effect the cms.

siinghd:
merging for now 

/bounty $20

paahaad:
> @paahaad hi thanks for point out this, after commenting out does the thumbnails and everything as before?

Yes. I have tested. it will not effect the cms.

PR #1212, opened 2024-09-13
Issue: #1211

### LINKED ISSUE #1211: bug: on Login Page, Phone number login displays email suggestions and blocks login without an email
**Describe the bug**
Login form should allow users to sign in using either their phone number or email. However, when a phone number is entered, the form still displays email suggestions like "9090900904@gmail.com," and prevents login without entering an email. I have to disable these email suggestions when a valid phone number is detected.

**To Reproduce**
Users should be able to log in using either an email or a phone number.
When a phone number (e.g., "9090900904") is entered, no email suggestions should appear, and the form should process the login correctly.

**Expected behavior**
Login using phone number also

**Actual Behavior**
When entering a phone number, the form incorrectly displays email suggestions, such as "9090900904@gmail.com."
The login attempt is blocked, as it expects an email instead of a phone number.

**Steps to Reproduce**
Implement the login form with fields for email and password.
Enter a valid 10-digit phone number into the email field.
Observe that email suggestions are generated (e.g., "9090900904@gmail.com").
Try to log in with the phone number — it fails without entering a valid email address.

**Screenshots or GIFs**
Screenshot 2024-09-14 at 1 37 58 AM

kingpranav21:
### Description
This pull request addresses issues with the email domain suggestions component:

- **Clear suggestions on mismatch**: If the user types a phone number or a custom domain that does not match the preset domains, the suggestions will disappear.
- **Hide suggestions on blur**: If the user doesn’t select a suggestion and clicks outside the input field, the suggestions will be cleared.

### PR Fixes:
- 1 Updated `handleEmailChange` to clear suggestions for phone numbers and mismatched domains.
- 2 Added `onBlur` event handler to clear suggestions when the user clicks outside the input field.

Resolves #1211 

### DIFF
--- a/src/components/Signin.tsx
+++ b/src/components/Signin.tsx
@@ -46,6 +46,13 @@ const Signin = () => {
       ...prevState,
       emailReq: false,
     }));
+    
+    // Check if the input is a phone number
+    const phoneNumberRegex = /^[0-9]{10}$/;
+    if (phoneNumberRegex.test(value)) {
+      setSuggestedDomains([]); // Clear suggestions for phone numbers
+      return;
+    }
 
     if (!value.includes('@')) {
       setSuggestedDomains(emailDomains);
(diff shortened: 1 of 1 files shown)


devsargam:
Thank you sir

PR #1232, opened 2024-09-15
fix bookmark and view of bookmark

### LINKED ISSUE #1231: bug: Bookmark is broken.
**Describe the bug**
1. duplicates records are being generated in bookmark table, due to incorrect flow of bookmark data to course page.
2. No support for notion docs or folder to bookmark.

[Screencast from 15-09-24 09:50:55 PM IST.webm](https://github.com/user-attachments/assets/f8be8bb2-f9d0-47c4-b4ba-78a5b5b1b55d)

paahaad:
### PR Fixes:
- 1 Added support for folder and notion doc
- 2 fix the bookmark state in the course view page. which was causing the bookmark button be active thus allowing the user to create same bookmark again polluting the db. 

Resolves #1231  
[Screencast from 15-09-24 09:52:22 PM IST.webm](https://github.com/user-attachments/assets/e9cebde8-597c-4c53-8811-14c7c2164776)

### DIFF
--- a/src/components/bookmark/BookmarkList.tsx
+++ b/src/components/bookmark/BookmarkList.tsx
@@ -21,12 +21,28 @@ const BookmarkList = ({
         {bookmarkData.map((bookmark) => {
           const {
             contentId,
-            content: { type, parent, title, hidden, thumbnail },
+            content: { type, parent, title, hidden, thumbnail, courses },
           } = bookmark;
-          if (type === 'video' && parent && !hidden) {
+          if ((type === 'video' || type === 'notion') && parent && !hidden) {
             const { id: folderId, courses } = parent;
             const courseId = courses[0].courseId;
             const videoUrl = `/courses/${courseId}/${folderId}/${contentId}`;
+
+            return (
+              <ContentCard
+                type={type}
+                key={contentId}
+                title={title}
+                image={thumbnail || ''}
+                onClick={() => {
+                  router.push(videoUrl);
+                }}
+                bookmark={bookmark}
+                contentId={contentId}
+              />
+            );
+          } else if (type === 'folder' && !parent) {
+            const videoUrl = `/courses/${courses[0].courseId}/${courses[0].contentId}`;
 
             return (
               <ContentCard
--- a/src/actions/bookmark/types.ts
+++ b/src/actions/bookmark/types.ts
@@ -17,5 +17,6 @@ export type ReturnTypeDeleteBookmark = ActionState<
 export type TBookmarkWithContent = Bookmark & {
   content: Content & {
     parent: { id: number; courses: CourseContent[] } | null;
+    courses: CourseContent[];
   };
 };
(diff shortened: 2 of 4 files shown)


hkirat:
@devsargam

src/components/bookmark/BookmarkList.tsx
```diff
@@ -21,13 +21,29 @@ const BookmarkList = ({
...
+                title={title}
+                image={thumbnail || ''}
+                onClick={() => {
+                  router.push(videoUrl);
+                }}
+                bookmark={bookmark}
+                contentId={contentId}
+              />
+            );
+          } else if (type === 'folder' && !parent) {
```
siinghd, line 44: why this else sir?  from what i can see you are returning the same thing

paahaad: videoUrl is diff in both the case. there was already an implementation for type == 'video'. I do not want to break that. plus control will not go inside if in case of folder because parent will be null.

paahaad:
@siinghd let me know if you feel some changes is required. 

also, rn in production user can create same bookmark multiple time. creating multiple entry of same contentId. I think we should change the schema of bookmark to 

```
model Bookmark {
  id        Int      @id @default(autoincrement())
  userId    String
  contentId Int
  user      User     @relation(fields: [userId], references: [id], onDelete: Cascade)
  content   Content  @relation(fields: [contentId], references: [id])
  createdAt DateTime @default(now())

  @@unique([userId, contentId])
} 
```
here is a script to delete multiple records then we can do prisma migrate.

```
import { PrismaClient } from '@prisma/client'

const prisma = new PrismaClient()

async function deduplicateBookmarks() {
  console.log('Starting deduplication process...')

  const allBookmarks = await prisma.bookmark.findMany({
    select: {
      id: true,
      userId: true,
      contentId: true,
    },
    orderBy: {
      createdAt: 'asc',
    },
  })

  console.log(`Total bookmarks found: ${allBookmarks.length}`)

  const uniqueCombinations = new Set()
  const duplicateIds = []

  for (const bookmark of allBookmarks) {
    const key = `${bookmark.userId}-${bookmark.contentId}`
    if (uniqueCombinations.has(key)) {
      duplicateIds.push(bookmark.id)
    } else {
      uniqueCombinations.add(key)
    }
  }

  console.log(`Duplicate bookmarks found: ${duplicateIds.length}`)

  if (duplicateIds.length > 0) {
    const deleteResult = await prisma.bookmark.deleteMany({
      where: {
        id: {
          in: duplicateIds,
        },
      },
    })

    console.log(`Deleted ${deleteResult.count} duplicate bookmarks`)
  } else {
    console.log('No duplicates found')
  }

  console.log('Deduplication process completed')
}

deduplicateBookmarks()
  
  ```

devsargam (approved):

devsargam:
thankss

PR #1253, opened 2024-09-18
resolved issue : 1252

### LINKED ISSUE #1252: bug: Content Overflow on Header and Theme Icon Redundancy
Description:
There are two primary issues observed in the current UI:

Content Overflow on Header:

The content pane (on the right side, "Course Content") is overlapping onto the header section when expanded. This creates a visual disruption as the content is not contained within its own section and instead bleeds over onto the header.

Redundant Change Theme Icon:
The "Change Theme" icon is currently placed both in the profile section and the header. Since the header already contains a theme toggle option, the one in the profile section is redundant and should be removed to avoid confusion and streamline the UI.

image
image

wraith2009:
### PR Fixes:
- 1 resolved and introduced overflow-y inside the div it was needed 
- 2 removed change theme btn from profile as it is already present in header.

Resolves #1252 

ps: i only changes 2 files idk how it is showing changes in 17files , sry its my first time facing this although all the files are same but it shows 17 files changes. 
 

### DIFF
--- a/src/app/admin/add-course/page.tsx
+++ b/src/app/admin/add-course/page.tsx
@@ -1,12 +1,7 @@
 'use client';
 
 import { Button } from '@/components/ui/button';
-import {
-  Card,
-  CardContent,
-  CardHeader,
-  CardTitle,
-} from '@/components/ui/card';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import {
   Form,
   FormControl,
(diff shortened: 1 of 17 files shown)


karranx:
the files that are changed automatically is because you have different linting rules from the projects which causes changes in indentation, single/double quotes etc. make sure to follow the linting rules of the project otherwise the PR will fail the automatic checks and wont be merged.

PR #1256, opened 2024-09-18
Bug Fix for Login Input Field Auto-Suggestions and Phone Number Login Issue

SargunSinghSethi:
### **Description**
I encountered a significant bug in the login functionality of the CMS system, where users were initially allowed to log in using their phone number and password. This issue arises because of the email input field’s auto-suggestion feature, which interferes with the login process, particularly when users enter their phone number.

### **Video Solution**
https://github.com/user-attachments/assets/983e8390-b4e8-4e01-aa69-821268ad4fd8

### **Problem Overview:**

1. Initial System Behavior:

- Users could log in either using an email and password or a phone number and password. However, many users initially registered with a phone number.
- Around 33% of users, based on initial records, were logging in using their phone numbers instead of an email.

2. **Current Issue:**

- The email input field is now giving auto-suggestions based on whatever the user types, even if it's a phone number. For example, if someone types "999", it suggests "999@gmail.com" and other similar options, even though it's a phone number.
- On mobile devices, these suggestions often overlap with the submit button, making it impossible for users to log in.

3. Solution Implemented: To address this, I modified the behavior of the auto-suggestions feature in the email input field.

- New Logic: Suggestions will only appear if the string entered contains the "@" symbol. If no "@" is present, the system will assume it's a phone number and prevent suggestions from appearing.
- This fix ensures that when users type a phone number (e.g., "999"), no unnecessary email suggestions are shown, and users can log in smoothly.

**Why This is Critical: This fix is crucial because:**

- A significant percentage of users (about 33%) rely on phone number login.
- On mobile devices, the auto-suggestion overlay blocks critical UI elements, preventing users from completing the login process.

@hkirat @devsargam @siinghd @TanmayDhobale

### DIFF
--- a/src/app/admin/add-course/page.tsx
+++ b/src/app/admin/add-course/page.tsx
@@ -1,12 +1,7 @@
 'use client';
 
 import { Button } from '@/components/ui/button';
-import {
-  Card,
-  CardContent,
-  CardHeader,
-  CardTitle,
-} from '@/components/ui/card';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
 import {
   Form,
   FormControl,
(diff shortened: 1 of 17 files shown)


devsargam:
Doesn't matter much sir. We can ignore this for now.

PR #1288, opened 2024-09-20
Fix : Improved the UI mouse changed to cursor-pointer when hovered

GARY121github:


devsargam:
There's a lot of noise in this pr. Can u reduce the diff???

PR #1290, opened 2024-09-20
ENH:  Leveage T3 ENV Type safety

vanshavenger:
Moved the Env varibales to the t3-oss/t3-env for better env typesafety.

Few issues i found down the line.
1. The CMS variable always comes as a string in the process.env so statements like if(process.env.LOCAL_CMS_PROVIDER) won't work properly. In this i have converted the env cariable to Boolean to fix this issue. 
2. Added ENV variables which were missing in the env.sample but was used in the repo

There was not linting and format error as you can see in the attach image the build was successful. 

If any changes required, please let me know to fix them on priority

Screenshot 2024-09-21 at 2 20 27 AM


devsargam:
Don't think it's needed here atm. It creates a lot of complexity too. I really love t3 env but closing this. Sorry

Issue #1314, opened 2024-09-25
bug: Theme changing button appears two times.


mortale2004:
The button for the changing theme from  dark mode to light or light mode to dark appears two times there is need to remove that button from the one place.
![image](https://github.com/user-attachments/assets/c6f96138-fe19-43c8-bb9b-a3d64e96cc36)

sampath24-ss:
Is the issue resolved?

mortale2004:
Solved

PR #1315, opened 2024-09-25
fix: The theme toggle button appears twice.

mortale2004:
### PR Fixes:
- 1 I have fixed the theme toggle button which appears two times by removing from the dropdown.
- 2 Formatted the all project to fix the formatting issue.

The issue was #1314


devsargam:
Can u fix conflicts. Also provide some good screenshots

mortale2004:
Sure why not

mortale2004:
Here is the screenshot image

devsargam (approved):

devsargam:
Thank you sir

PR #1325, opened 2024-09-25
Fix new UI bug

### LINKED ISSUE #1180: bug: new ui bug
Screenshot 2024-09-10 at 10 20 03 PM

shaurya35:
### PR Fixes:
- Fixed the issue where the video player was taking so much space that its part was being cut when on higher resolution

### Old UI
Screenshot 2024-09-19 035109

### New UI
Screenshot 2024-09-19 034848

### Responsiveness?
Screenshot 2024-09-19 034930

### What does it Resolve?
- For devices with a width greater than 1280px, the path and the button are on the same line.
- For all others, it's the same as it is.

### What is changed?
- For 'xl' size and above, the CSS property changed to absolute and overlapped the button component.
- Made the z-index of button 50 to it to work.
- Tested in many scenarios.

_### Open for any other Suggestions or Changes_

Resolves #1180 

### DIFF
--- a/src/components/CourseView.tsx
+++ b/src/components/CourseView.tsx
@@ -38,8 +38,8 @@ export const CourseView = ({
     ? 'folder'
     : courseContent?.value.type;
   return (
-    <div className="flex w-full flex-col gap-8 pb-16 pt-8">
-      <div className="flex flex-col gap-4">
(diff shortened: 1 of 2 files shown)


devsargam:
Thank you sir

PR #1349, opened 2024-09-29
IP-based JWT validation and session management

### LINKED ISSUE #1156: bug: Store ips of the user in the db along with token, make sure token is valid for that ip

TanmayDhobale:
This PR addresses the security issue related to stolen JWTs and unauthorized dashboard access. It implements IP-based token validation and enhances overall security measures.

Key changes:
1. Store and validate IP addresses with JWTs
2. Implement short-lived access tokens (15 minutes) and refresh token mechanism
3. Log and limit IP changes to prevent abuse
4. Add rate limiting to protect against brute force attacks
5. Implement security headers for better protection against common web vulnerabilities

Resolves #1156


TanmayDhobale:
@devsargam can u check this

devsargam:
I am unsure that this is a good thing to add

devsargam:
few problems that I can see right now:
- public ip address of users keep on rotating (example: when switching wifis or mobile data)
- this will lot all users out when it is implemented

TanmayDhobale:
> few problems that I can see right now:
> 
> * public ip address of users keep on rotating (example: when switching wifis or mobile data)
> * this will lot all users out when it is implemented

yea could lead to unnecessary logouts

TanmayDhobale:
> I am unsure that this is a good thing to add

Yeah, but @hkirat  raised an issue about this.

PR #1374, opened 2024-10-02
FIX: Typescript error for react 17 or above

vanshavenger:
### PR Fixes:
Screenshot 2024-10-02 at 8 20 02 PM

In react 17 or above the follwoing change can help to not use React import in each file other wise it shows a warning as you can see in the screenshot attached


vanshavenger:
Better solution is to have React imports in those files, After reaearching more

Should i add imports to all the files to resolve the typescript error.

devsargam:
Thank you sir

PR #1376, opened 2024-10-02
Fix: Husky, Pre-commit 

vanshavenger:
The following PR fixes all the problems. that husky was facing and the prettier and lint errors that were conflicting on the schema .ts file

in schema.ts file earlier the wrap regex was causign lint erros as prettier remove the parens, so instead move it to New RegExp which is inbuilt and can be useful 

Formatted all the files 

Added more checks to ensure code standars are met
this PR also format all files to give you a brief how the formatting is done

and important to ensure all the Code standards are met and fulfilled.
Screenshot 2024-10-02 at 9 00 28 PM

Husky/pre-commit also got revamped to ensure more strict checking


devsargam:
There are many file changes in your pr just had to close this since it will affect a lot of other people's prs

PR #1415, opened 2024-10-04
Feature/forget password

### LINKED ISSUE #1386: feature: forget password
I'm very frustrated when it ask to password and i signup with otp in cohortharkirat there is no option in forget password in 100xdevs and also when i want to reset password in cohortharkirat otp doesn't send in my email account.

atul24112001:
### PR Fixes:
- 1 Added forget password feature it will change password for both classx and cms
- 2 Fixed build & type Errors

Resolves #1386 


atul24112001:
Hey @devsargam please have a look

Important files to check for
src/app/api/user/change-password/route.ts
src/app/api/user/change-password/verify-otp/route.ts
src/components/Signin/index.tsx
src/components/Signin/SigninForm.tsx
src/components/Signin/ForgetPasswordForm.tsx

atul24112001:
https://github.com/user-attachments/assets/a917f209-1e21-4dc0-8961-9ad93cca6a1c

devsargam:
The problem is
- we fetch data from harkirat.classx website but we can not change the data there
- the database you are looking is just used for catching and not the true source

Problem with this approach is it will change password on app.100xdevs.com but not on other website

atul24112001:
@devsargam you are correct we have to change it at two place that’s what I am doing, if you check I am changing the password on harkirat.classes first, The OTP is sent by classx I have tracked their forgot password flow apis and integrated in cms backend once classx verifies the OTP and changes the password I am changing it for cms as well.

devsargam:
Just curious how did you get the appx headers? Also, I need to check this works on server side or not as well

atul24112001:
I have tracked the api and in code its already mentioned in authOptions file.
Screenshot 2024-10-06 at 11 59 51 AM

PR #1176, opened 2024-09-10
feat/search: Added search bar and other minor ui fix

ItsFlash10:
### PR Fixes:
- Added search bar back in the courses page and moved the siederbar trigger to the right since the sidebar is on the right side.
- Fixed the siderbar UI

Resolves #1175

@siinghd @devsargam @hkirat

https://github.com/user-attachments/assets/bb107666-603f-4d85-8646-4bd6c3989fe1

### Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] I assure there is no similar/duplicate pull request regarding same issue

PR #1427, opened 2024-10-04
Revert "feat/search: Added search bar and other minor ui fix"

devsargam:
Reverts code100x/cms#1176


src/app/api/search/route.ts
```diff
@@ -42,19 +39,6 @@ export async function GET(request: NextRequest) {
       where: {
         type: 'video',
         hidden: false,
-        parent: {
-          courses: {
-            some: {
-              course: {
-                purchasedBy: {
-                  some: {
-                    userId: session?.user?.id,
```
devsargam, line 51: @ItsFlash10 you set the cache to a particular user who opens it. This is wrong have to revert

PR #1430, opened 2024-10-05
Remove: Husky

vanshavenger:
Remove Husky globally

also move two @types to the dev dependencies from dependencies


vanshavenger:
Resolved

PR #1435, opened 2024-10-05
Added a shadow to the PostCard component.

Karthik150502:
Added a shadow to the PostCard component, so that the card has visible defined boundary in light theme.

### PR Fixes:
- 1 - Added some shadow to the PostCard component, for a better visibility in light theme.

Screenshot 2024-10-05 193832


Karthik150502:
The above image is the old one, the changed one looks like this;

image
@devsargam

devsargam:
Thanks

PR #1467, opened 2024-10-10
ui: Adjusted video layout when Sidebar opens

gupta-soham:
- Implements a `SidebarProvider` for managing Sidebar state and transitions
- Implements smooth transitions using `framer-motion`
- Adjusts video layout automatically when the sidebar opens

## Before

Before

## After

https://github.com/user-attachments/assets/29d363de-bf56-4d36-bdee-60c37fc26d99


devsargam:
Always the secondary sidebar from the right is meant to be floating over the components. Thankyou for the effort

PR #1482, opened 2024-10-16
better ui, reduced video frame such that it becomes easy to navigate …

NalinDalal:
…with video controls

### PR Fixes:
- 1
- 2


NalinDalal:
lemme know if there are any issues

PR #1483, opened 2024-10-16
Added scroll up button in notes

rohhann12:
### PR Fixes:
- 1. Added a scroll up button in notes so that the user doesn't have to keep scrolling up after he/she have read through them,
added video to show how efficient is for the user to scroll back up to `Show Content` button

fix-
https://github.com/user-attachments/assets/6de52f02-aa53-46c1-9954-a19e83218749

*tried to solve this using script from next/script and not using hooks to avoid usage of "use client"

- [] I have performed a self-review of my code
- [] I assure there is no similar/duplicate pull request regarding same issue

### DIFF
--- a/src/components/comment/Comments.tsx
+++ b/src/components/comment/Comments.tsx
@@ -2,9 +2,9 @@ import React from 'react';
 import { Avatar, AvatarFallback } from '../ui/avatar';
 import { TabType, QueryParams, ROLES } from '@/actions/types';
 import {
-  constructCommentPrismaQuery,
-  getUpdatedUrl,
-  paginationData,
+	constructCommentPrismaQuery,
+	getUpdatedUrl,
+	paginationData,
 } from '@/lib/utils';
 import CommentInputForm from './CommentInputForm';
 import { getComments } from '../../actions/comment/index';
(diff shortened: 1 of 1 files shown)


src/components/comment/Comments.tsx
```diff
@@ -83,7 +96,7 @@ const Comments = async ({
         )}
         <div className="flex flex-col justify-between gap-4 md:flex-row md:items-center">
           <h2
-            className={`text-xl font-bold tracking-tighter text-primary md:text-2xl`}
+            className={`text-xl font-bold taracking-tighter text-primary md:text-2xl`}
```
rohhann12, line 99: used gpt to format it correctly, dont know why it changed this, changing

src/components/comment/Comments.tsx
```diff
@@ -317,7 +330,29 @@ const Comments = async ({
           ))}
         </div>
       </div>
-      <div>{/* <Pagination dataLength={data.comments.length} /> */}</div>
+      <BackToTopButton />
+      <Script id="back-to-top-script" strategy="afterInteractive">
+        {`
+          function toggleBackToTopButton() {
```
devsargam, line 336: Do this function the react way not like this

devsargam (changes requested):

rohhann12:
too clumsy, putting up a new one, with minimal changes

PR #1521, opened 2024-10-27
Feat: Add autofocus functionality to the Question's search box and ad…

i-amansharma07:
### PR Fixes:
- Added autoFocus prop to the Search Component so that when some page needs a searchbox with auto focus it can pass **atoFocus={true}**
- Added  autoFocus to the Questions Page so that students can directly search for the keyword they are looking for which saves an extra click
- Added gap between search box and submit button which now looks cool btw


devsargam:
We already have cmdk which is much easier and better don't think this is necessary also this might bug with video palyers seek functionality closing

PR #1522, opened 2024-10-28
fix the endpoint for getting folders of course

harsimran-d:
### PR Fixes:
- 1 Make the logic to check if the user has bought a course same as the request that fetches all courses at /courses
- 2 Make the logic to retrieve all folders by courses and match the id there and not match with the content id


harsimran-d:
Hi @devsargam I think i have found the issue for one of the endpoint 
could you please review and see if it works.
thanks

devsargam:
hey thanks for raising the pr

devsargam:
ref: https://github.com/code100x/cms/blob/main/src/utiles/appx.ts

PR #1555, opened 2024-11-07
Optimized Docker setup for production

billuop123:
### PR Fixes:
- 1 Split the build and run stages to create a cleaner and more efficient image.
- 2 Set up a non-root user for running the application in the production stage, improving security.
-3 Moved dependency installation to the build stage, only copying the essential files into the production image to reduce size.
-4 Used DATABASE_URL as an ARG to generate the Prisma client only once in the build stage, making it production-ready.
Resolves #[1554] 

### DIFF
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -1,31 +1,33 @@
+# Build stage
 FROM node:20-alpine AS build
 
 WORKDIR /usr/src/app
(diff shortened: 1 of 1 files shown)


geekyharsh05:
@billuop123 Can you share the screenshot of the image size of the production image.

PR #1568, opened 2024-11-17
fix: resolve issue with right arrow keystroke functionality

thinley4:
### Before
The right arrow key stops functioning after being pressed three times.

[before.webm](https://github.com/user-attachments/assets/b4742406-5dca-4085-bb4e-ef6f0eb78b22)

### After

[after.webm](https://github.com/user-attachments/assets/7286a271-d0ad-4a2b-afe7-874a052d681e)


devsargam:
Thank you sir

PR #1580, opened 2024-11-24
feat: Custom Encrypted Player 

mehul-srivastava:
### PR Fixes:
- Implemented a custom encryption layer on top of hls.js specifically. The quality remains top-notch and the manifest is heavily encrypted avoiding piracy issues.

THIS IS JUST STEP 1 OF SOLVING PIRACY PROBLEM - I am currently working on my other ideas to make it harder for people to reverse engineer it.

https://github.com/user-attachments/assets/57065f88-cacc-430b-a874-6f4663aa8862

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "@discordjs/next": "^0.1.1-dev.1673526225-a580768.0",
     "@hookform/resolvers": "^3.6.0",
     "@icons-pack/react-simple-icons": "^9.4.0",
+    "@mehul-srivastava/hls.js": "^1.0.0",
     "@prisma/client": "^5.18.0",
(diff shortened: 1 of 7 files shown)


hkirat:
Creating our in house encryption might not be on that charts atm

mehul-srivastava:
> Creating our in house encryption might not be on that charts atm

Well it's not actually an in-house encryption from scratch. I just inherited what was already working on app.100x and integrated my solution on top of it. So everything works out of the box - but with high quality video and encrypted manifest (to avoid piracy issue)

devsargam:
Let's not go this direction since we are using the one provided by appx. This implementation is great but we do want to use the iframe

PR #1581, opened 2024-11-24
Continue watching button added to the course folders

Pratik610:
### Continue watching button

https://github.com/user-attachments/assets/a1f78a08-5185-45b5-8981-504d9db45eac


devsargam:
We already have the watch history thing. I don't think this is much needed since it solves similar problem

Pratik610:
fair enough

Pratik610:
> We already have the watch history thing. I don't think this is much needed since it solves similar problem

fair enough

PR #1612, opened 2024-12-05
Fix App Bar Icon Shrinking and Wrapper Overlapping with main contents Issues

rajveeerr:
### PR Fixes:
- Fixed app bar icon sizing by setting the min-width to 3.7rem (previously, the width was set to 4vw with no min-width), leading to  icons shrinking to an unusually small size.
- Increased padding in the wrapper to resolve the overlapping issue between the app bar and buttons.  

### Before and After Screenshots  

| **State**   | **Screenshot**          |  
|-------------|-------------------------|  
| **Before**  | Screenshot 2024-12-06 at 6 48 34 PM |  
| **After**   | Screenshot 2024-12-06 at 6 47 07 PM |  

### Checklist Before Requesting a Review
- [x] I have performed a self-review of my code.  
- [x] I assure there is no similar/duplicate pull request regarding the same issue.  

@devsargam @siinghd

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -216,7 +216,7 @@
 
 @layer utilities {
   .wrapper {
-    @apply h-full w-full max-w-7xl px-4 py-4 md:px-12 lg:pl-20 lg:pr-12 lg:mx-auto xl:max-w-[90rem];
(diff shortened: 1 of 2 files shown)


devsargam:
thanks

PR #1623, opened 2024-12-07
feat: update lastLoginIp saving on login

### LINKED ISSUE #1613: Store IP during signups, restrict tokens to IPs
right now tokens can be shared and multiple people can login
Restricting by IP should help in protecting that

Smnthjm08:
PR Fixes:
- Save user IP address on login

I have attempted to implement the functionality to save the user's IP address when they log in. While I believe the implementation works, I'm not entirely confident about it, and would appreciate your feedback and suggestions for improvement.

Resolves #1613

Checklist before requesting a review
 [x] I have performed a self-review of my code
 [x] I assure there is no similar/duplicate pull request regarding the same issue


Smnthjm08:
@hkirat @devsargam

Issue #1626, opened 2024-12-07
bug: something is up with week 7


hkirat:
<img width="950" alt="Screenshot 2024-12-07 at 9 42 40 PM" src="https://github.com/user-attachments/assets/a2ce7dd1-43b1-41be-8cbe-62401a826d54">

saurabhraghuvanshii:
@hkirat  it is because of new pr merged #1486

IkramBagban:
Hey @hkirat @saurabhraghuvanshii, could you please explain what this issue is about? I am not able to understand it. Is it about the title not showing? I think someone has already raised a PR for this.

saurabhraghuvanshii:
@IkramBagban ya I see it is resolve in #1622 . check it sir @hkirat

IkramBagban:
> @IkramBagban ya I see it is resolve in #1622 . check it sir @hkirat

But he mentioned that something is up with week 7.

srv-18:
@IkramBagban inside contents are not displayed in the sidebar of the show content.
![Screenshot 2024-12-11 181527](https://github.com/user-attachments/assets/5ad973bd-2cf2-4346-be97-ee5d2cc47414)

IkramBagban:
[...]

PR #1648, opened 2024-12-10
fixed content title not visible in sidebar

iamlikeaproforyou:
Fixed the content title not visible in sidebar also added cursor pointer to the checkbox
Before
image

After
image

fixes: #1626

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -191,7 +191,7 @@ export function Sidebar({
             </AccordionItem>
           );
         }
-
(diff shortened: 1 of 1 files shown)


Mh-Alim:
@hkirat bug is still there

handlerFive:
@devsargam - I think the fix is not yet deployed as the issue is still there

iamlikeaproforyou:
@Mh-Alim It has been fixed in the main branch yet to get deployed to the production branch

PR #1705, opened 2025-01-09
added docker optimised docker file with better cacheing

rahul48635:
added some steps in docker file for optimising cheching

### PR Fixes:
- 1
- 2


devsargam:
Docker file is already optimized with `Dockerfile.prod`

PR #1712, opened 2025-01-14
feat: Implemented form for reporting issues encountered while accessing the course

### LINKED ISSUE #1711: feature: add a form to report issues while accessing courses
Yes, the issue is that when users encounter problems accessing the course, the current "Contact Us" link uses a Link tag that opens the browser's tab which is blank. However, this process is not ideal, as it can be disruptive. This makes it difficult for users to report issues efficiently.

I have created a form component that allows users to submit issues they face while accessing a course. The form includes fields like first name, last name, email and message, which users can fill out to report the problem. When the form is submitted, the Nodemailer package is used to send an email directly to the support team, ensuring that issues are reported in an organized and consistent manner. This solution eliminates the disruption caused by the Link tag and provides a more user-friendly way to contact support.

Current Issue: 
https://github.com/user-attachments/assets/d011c092-3d62-4915-9ce9-6ea62ec84a5b

ToseebNadaf:
### PR Fixes: 
- This PR updates the contact form in the application to allow users to report issues.
- The email sent now includes the user's name, email, and their issue description. 
- Updated the backend endpoint to handle issue reporting.
- Enhanced the `sendMail` function to include full user details.
- Added error handling for missing fields in the form.

Note: 
--- Just create app password from manage google and add it to the env file

Resolves #1711 

- [✅] I have performed a self-review of my code
- [✅] I assure there is no similar/duplicate pull request regarding same issue

Before: 

https://github.com/user-attachments/assets/ade26763-29e7-4074-bd74-c5b8896b3a8e

After: 

https://github.com/user-attachments/assets/ab33169b-2134-44ca-8f55-fd76f42a4cbe


ToseebNadaf:
@devsargam and @hkirat please look into this and suggest me any changes if needed

src/lib/nodemailer.ts
```diff
@@ -0,0 +1,31 @@
+import nodemailer from "nodemailer";
+
+const transporter = nodemailer.createTransport({
```
devsargam, line 3: FYI, gmail limits the number of emails that we can send over a day. So, we might reach that limit very easily

src/app/api/send-email/route.ts
```diff
@@ -0,0 +1,28 @@
+import { NextResponse } from "next/server";
+import { sendMail } from "@/lib/nodemailer";
+
+export async function POST(req: Request) {
```
devsargam, line 4: Missing authentication check here

devsargam (changes requested):

devsargam:
I will check with the team if this feature is needed or not and will update

src/lib/nodemailer.ts
ToseebNadaf: Yes, but Gmail enforces daily email limits regardless of the method used (SMTP, web interface, or API). If we’re likely to hit these limits, then we have to switch to a scalable service like SendGrid for handling higher email volumes effectively.

PR #1728, opened 2025-01-29
Added appropriate message for filter #1725

### LINKED ISSUE #1725: feature: Proper display of message in UI when any filter result is null
-> I tried to apply the filter (suppose Watched) inside (say week 5) and the result was null as I haven't watched any and the screen appeared black screen making the User Experience bad. 

-> When the result is null from the filter, display some message like "No watched content found in this section" or anything similar to it which will make a good user experience.

**Additional context**
Screenshot

Image

Vinit1014:
### PR Fixes:
- 1 A short and simple message displayed when filtered content is null to make a better user experience.
- 2 Displays different message for different filter.

Resolves #1725 

Screenshot 2025-01-29 143849
Screenshot 2025-01-29 143913

### DIFF
--- a/src/components/FolderView.tsx
+++ b/src/components/FolderView.tsx
@@ -36,6 +36,23 @@ export const FolderView = ({
     currentfilter,
   );
 
+  if (filteredCourseContent?.length === 0) {
+    const filterMessages: Record<string, string> = {
+      watched: "You haven't completed any content in this section yet.",
+      watching: "No content currently in progress.",
+      unwatched: "No new content available to watch.",
+      all: "No content available in this section.",
+    };
+  
+    return (
+      <div className="mt-56 flex">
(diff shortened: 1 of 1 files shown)


src/components/FolderView.tsx
```diff
@@ -36,6 +36,23 @@ export const FolderView = ({
     currentfilter,
   );
 
+  if (filteredCourseContent?.length === 0) {
+    const filterMessages: Record<string, string> = {
```
devsargam, line 40: Best way to write typescript is to avoid writing types and letting the compiler decide

```suggestion
    const filterMessages = {
```

devsargam (changes requested):

devsargam:
Thank you appreciate it

Vinit1014:
Welcome sir

PR #1730, opened 2025-02-03
Updated AppxVideoPlayer.tsx - Added full screen feature on double click

adityasharmawork:
### PR Fixes:

I was facing some issues, and, wanted a feature, to make the video go to full screen on double click, just like YouTube, as, clicking on that little icon multiple times was quite annoying. So, I have added this feature to make video go full screen on double click.

### DIFF
--- a/src/components/AppxVideoPlayer.tsx
+++ b/src/components/AppxVideoPlayer.tsx
@@ -40,5 +40,16 @@ export const AppxVideoPlayer = ({
     src={url}
     className="w-full rounded-lg aspect-video"
     allowFullScreen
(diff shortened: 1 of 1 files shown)


devsargam (commented):
This won't work on iframe. It only does in video element. If I am wrong feel free to raise another pr

adityasharmawork:
It was working on my machine. Sorry, I became a bit late in recording the video, as, I was building some projects and especially going through some codebases for GSoC.

Will try it out again, if it works, and, doesn't cause any issue, I will create another PR.
Thanks for replying btw.

PR #1761, opened 2025-02-22
fixed issue #1755:video starts from where it was left

### LINKED ISSUE #1755: bug: Video starts to play in low resolution and ..
**Describe the bug**
when I start playing a recorded class it automatically sets resolution to 360p 
when site reloads video begins from starting 

**To Reproduce**
1. buy course then try watching a video
2. reload the site and see

**Expected behavior**
Its great we can set different resolutions for a video but since below 1080p the video looks awful and blurred.
when initial resolution of the video should be 1080p

on reload the  video should play back from same point as we left it  

**Screenshots or GIFs**

https://github.com/user-attachments/assets/7fe3ce79-33a9-4433-9083-7179d3605af4

 - firefox
 - 132.0 (64 bit)

RatnadeepRoy01:
### PR Fixes:
- 1
- 2

Resolves #1755 

### DIFF
--- a/src/components/VideoPlayer2.tsx
+++ b/src/components/VideoPlayer2.tsx
@@ -17,6 +17,7 @@ import { toast } from 'sonner';
 import { createRoot } from 'react-dom/client';
 import { PictureInPicture2 } from 'lucide-react';
 import { AppxVideoPlayer } from './AppxVideoPlayer';
+import { loadESLint } from 'eslint';
 
 // todo correct types
 interface VideoPlayerProps {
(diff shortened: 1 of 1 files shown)


RatnadeepRoy01:
The progress of the video is saved on localstorage and retrieved from the storage each time the page mounts again

RatnadeepRoy01:
The progress of the video is saved on localstorage and retrieved from the storage each time the page mounts again

RatnadeepRoy01:
The progress of the video is saved on localstorage and retrieved from localstorage when the page mounts again

devsargam:
It's better to get that data from db only. Sync across devices and all. Also, makes it less complicated

RatnadeepRoy01:
Hi, i may be wrong but when page re mounts again it starts from the very
beginning may be retrieving from db is slow or not pushing to db at all
should I look into it

PR #1805, opened 2025-04-03
Implement Floating Navbar

DufferDeepu:
### PR Fixes:
- 1. Implemented a floating navbar: Changed the static navbar to a floating navbar that sticks at the top when scrolling.
- 2. Added required CSS : Updated styles for extra shadow for floating effect.

- I have tested the floating navbar functionality in different screen sizes (desktop/mobile).
- I have verified that no other part of the UI breaks due to these changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@uiw/react-md-editor": "^4.0.4",
     "autoprefixer": "^10.4.20",
     "axios": "^1.6.2",
-    "bcrypt": "^5.1.1",
+    "bcryptjs": "^3.0.2",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.0",
     "cmdk": "1.0.0",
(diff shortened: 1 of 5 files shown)


geekyharsh05:
@DufferDeepu Why are you using bcryptjs??

package.json
```diff
@@ -46,7 +46,7 @@
     "@uiw/react-md-editor": "^4.0.4",
     "autoprefixer": "^10.4.20",
     "axios": "^1.6.2",
-    "bcrypt": "^5.1.1",
+    "bcryptjs": "^3.0.2",
```
geekyharsh05, line 49: Why?

src/lib/auth.ts
```diff
@@ -1,7 +1,7 @@
 import db from '@/db';
 import CredentialsProvider from 'next-auth/providers/credentials';
 import { JWTPayload, SignJWT, importJWK } from 'jose';
-import bcrypt from 'bcrypt';
```
geekyharsh05, line 4: Switching to `bcryptjs` may break auth for users with`bcrypt` hashed passwords.

DufferDeepu: Thanks for pointing that out! I'll correct it.

PR #1847, opened 2025-06-09
errors occurred due to bcript after setting up locally and hence it i…

DebasishBarai:
### PR Fixes:
- ✅ Replaced `bcrypt` with `bcryptjs` to avoid native compilation issues during setup.
- ✅ Updated all relevant imports and usage in the codebase.
- ✅ Confirmed successful user authentication and registration after replacement.

### Resolves:
_No linked issue._

---

### ✅ Checklist before requesting a review


geekyharsh05:
@DebasishBarai IMO there will be backwards compatibility issue for the existing users!

Issue #1864, opened 2025-07-17
bug: Sidebar toggle’s bottom border is misaligned with the top navbar's horizontal line in both expanded and collapsed states.


x15sr71:
**Describe the bug**
The bottom border of the sidebar toggle button is misaligned with the top navbar's horizontal border. This happens in both expanded and collapsed sidebar states, causing a visual inconsistency.

**To Reproduce**
1. Start the app locally or visit the live site.

2. Observe the sidebar toggle section at the top-left.

3. Toggle between collapsed and expanded views.

4. Notice the border below the toggle doesn't align with the navbar's bottom border.

**Expected behavior**
The bottom border of the toggle section should line up perfectly with the navbar’s lower border to maintain consistent UI alignment.

**Screenshots or GIFs**

<img width="247" height="101" alt="Image" src="https://github.com/user-attachments/assets/c46efeb0-b244-4534-905c-5665ec5ea703" />
<br><br>
<img width="286" height="104" alt="Image" src="https://github.com/user-attachments/assets/6b94bc59-9f53-42fd-99d9-d24d87a66ec5" />

**Info (please complete the following information):**
 - Browser chrome
 - Version 138.0.7204.102

PR #1866, opened 2025-07-17
fix: align sidebar toggle bottom border with navbar (#1864)

x15sr71:
### PR Fixes:
1. Fixed incorrect padding for sidebar toggle to visually align with navbar border.
2. Ensured the fix works consistently in both collapsed and expanded sidebar states for visual consistency.

Resolves #[1864] 


x15sr71:
One CI test is failing because it runs `pnpm run format:check`, which requires `prettier-plugin-tailwindcss`. However, this plugin isn't listed as a dev dependency in the cms/package.json. Should we add it to devDependencies, or update the CI/config to avoid requiring it?

